PR #3, opened 2025-09-25
Cleanup: Removed dependencies, ran pre-commit with --run-all

arangatang:
*Description of changes:*

Removed three dependencies:
1. httpx_auth_awssigv4 - We should not send customer credentials to 3p dependency
2. requests_auth_aws_sigv4 - Unused
3. loguru - No need, we can use python built in logger

Further ran `pre-commit run --all-files` to get formatting correct.

*Testing*
Connected to a mcp and saw connection succesfull and i could list tools

### DIFF
--- a/src/aws_mcp_proxy/utils.py
+++ b/src/aws_mcp_proxy/utils.py
@@ -1,3 +1,17 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


src/aws_mcp_proxy/utils.py
```diff
@@ -1,21 +1,8 @@
-# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
JacqualinPotgieter, line 1: I see this snippet is in almost all files. Should we not keep it?

I am not sure about specifics regarding our license though.

arangatang: Yeah we must, tbh we should have a build step that checks every file for it. 
I'll add that in another PR. But for now i'll do a manual pass and add them where needed.

PR #10, opened 2025-09-29
Setup issue management

acmlau:
*Description of changes:*
- Setup issue management including issue templates for bugs, features, questions

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -86,7 +86,7 @@ body:
     id: search-verification
     attributes:
       label: Search verification
-      description: Please verify that you have completed the following:
+      description: Please verify that you have completed the following
       options:
         - label: I have searched existing issues and verified this is not a duplicate
           required: true


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,83 @@
...
+    id: operating-system
+    attributes:
+      label: Operating System
+    validations:
+      required: true
+
+  - type: input
+    id: server-version
+    attributes:
+      label: Server Version
```
arangatang: What is this? Would it be the URL?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,83 @@
...
+
+  - type: input
+    id: operating-system
+    attributes:
+      label: Operating System
+    validations:
+      required: true
+
+  - type: input
+    id: server-version
```
wzxxing: Should this be the proxy version?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,83 @@
+---
```
arangatang, line 1: We probably want to require submitters to verify there are no existing bug reports of the same bug. And if there are related but not the same ask them to reference them.

.github/ISSUE_TEMPLATE/bug_report.yml
acmlau: Ah correct would be better to change that to proxy version or more generic just "version"

wzxxing: just "version" sounds fine to me.

.github/ISSUE_TEMPLATE/bug_report.yml
acmlau: Added, refer to lines 85-100

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -67,11 +67,11 @@ body:
       required: true
 
   - type: input
-    id: server-version
+    id: version
     attributes:
-      label: Server Version
+      Release Version / Commit id
```
wzxxing: note that including commit ID requires some custom logic when we build and publish the package. Right now we do not have it, only version is available at the moment. 

If you are interested it, you can check for some available options.

acmlau: Will keep it to Release version then

arangatang: @wzxxing 
Thing is they could be running from source as well, so commit id is also useful imo

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,100 @@
...
+      description: |
+        e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. associated pull-request, stackoverflow, slack, etc
+    validations:
+      required: false
+
+  - type: checkboxes
+    id: search-verification
+    attributes:
+      label: Search verification
+      description: Please verify that you have completed the following:
```
arangatang:
```suggestion
      description: Please verify that you have completed the following
```

arangatang: I think the `:` causes the file to throw the error [here](https://github.com/acmlau/aws-mcp-proxy/blob/acmlau/.github/ISSUE_TEMPLATE/bug_report.yml)

PR #16, opened 2025-09-30
Add region to CLI arguments

acmlau:
*Description of changes:*
* Added `--region` argument to specify region from command line
* Added validation and determination for region
* Updated tests to reflect changes
* Moved environment variables fetching to main execution of script

### DIFF
--- a/aws_mcp_proxy/sigv4_helper.py
+++ b/aws_mcp_proxy/sigv4_helper.py
@@ -17,7 +17,6 @@
 import boto3
 import httpx
 import logging
-import os
 from botocore.auth import SigV4Auth
 from botocore.awsrequest import AWSRequest
 from botocore.credentials import Credentials
@@ -150,9 +149,7 @@ def create_aws_session(profile: Optional[str] = None) -> boto3.Session:
     return session
 
 
-def create_sigv4_auth(
-    service: str, region: str, profile: Optional[str] = None
-) -> SigV4HTTPXAuth:
+def create_sigv4_auth(service: str, region: str, profile: Optional[str] = None) -> SigV4HTTPXAuth:
     """Create SigV4 authentication for AWS requests.
 
     Args:
--- a/tests/unit/test_sigv4_helper.py
+++ b/tests/unit/test_sigv4_helper.py
@@ -245,7 +245,6 @@ class TestCreateSigv4Auth:
         assert result.region == 'test-region'  # default region
         assert result.credentials == mock_credentials
 
-
     @patch('aws_mcp_proxy.sigv4_helper.create_aws_session')
     def test_create_sigv4_auth_with_explicit_region(self, mock_create_session):
         """Test creating SigV4 auth with explicit region parameter."""
@@ -308,7 +307,9 @@ class TestCreateSigv4Client:
 
         # Test client creation with custom headers
         custom_headers = {'Custom-Header': 'custom-value'}
-        result = create_sigv4_client(service='test-service', region='test-region', headers=custom_headers)
+        result = create_sigv4_client(
+            service='test-service', region='test-region', headers=custom_headers
+        )
 
         # Verify client was created with merged headers
         call_args = mock_client_class.call_args
@@ -352,7 +353,10 @@ class TestCreateSigv4Client:
 
         # Test client creation with additional kwargs
         result = create_sigv4_client(
-            service='test-service', region='test-region', verify=False, proxies={'http': 'http://proxy:8080'}
+            service='test-service',
+            region='test-region',
+            verify=False,
+            proxies={'http': 'http://proxy:8080'},
         )
 
         # Verify client was created with additional kwargs


aws_mcp_proxy/server.py
```diff
@@ -47,8 +47,8 @@ async def setup_mcp_mode(mcp: FastMCP, args) -> None:
     service = determine_service_name(args.endpoint, args.service)
 
     # Get profile and region
-    profile = args.profile or os.getenv('AWS_PROFILE')
-    region = os.getenv('AWS_REGION', 'us-west-2')
+    profile = args.profile
+    region = args.profile
```
wzxxing: region?

acmlau: Good catch, updated with new commit

tests/test_sigv4_helper.py
```diff
@@ -243,7 +243,7 @@
         # Verify auth was created correctly
         assert isinstance(result, SigV4HTTPXAuth)
         assert result.service == 'test-service'
-        assert result.region == 'us-west-2'  # default region
+        assert result.region == 'us-east-1'  # default region
```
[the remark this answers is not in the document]
acmlau: Test were passing before because we have a fall back mechanism here for the region as well, and it was set to `us-west-2` before https://github.com/aws/aws-mcp-proxy/blob/main/aws_mcp_proxy/sigv4_helper.py#L174-L175

aws_mcp_proxy/sigv4_helper.py
```diff
@@ -172,7 +172,7 @@ def create_sigv4_auth(
 
     # Get region from parameter, environment variable, or default
     if not region:
-        region = os.environ.get('AWS_REGION', 'us-west-2')
```
arangatang, line 175: We should pass the region here, from the parsed args rather than fetching it again.

acmlau: As a standalone function, it makes sense to check whether region is not set. 
I did notice that we never passed the region as a parameter when creating the sigv4_auth https://github.com/aws/aws-mcp-proxy/blob/main/aws_mcp_proxy/server.py#L60, will create new PR to address that + removing some unused parameters

aws_mcp_proxy/utils.py
```diff
@@ -81,3 +90,35 @@ def determine_service_name(endpoint: str, service: Optional[str] = None) -> str:
...
+    """
+    if region:
+        return region
+
+    # Parse AWS region from endpoint URL
+    parsed = urlparse(endpoint)
+    hostname = parsed.hostname or ''
+
+    # Extract region name (pattern: service.region.api.aws or service-name.region.api.aws)
+    region_match = re.search(r'\.([a-z0-9-]+)\.api\.aws', hostname)
```
wzxxing, line 116: Could you also add the support for this type of hostnames?

```
https://bedrock-agentcore.us-west-2.amazonaws.com
```


```suggestion
    region_match = re.search(r'\.([a-z0-9-]+)\.(api\.aws|amazonaws\.com)$', hostname)
```

aws_mcp_proxy/sigv4_helper.py
wzxxing: By the time we create sigv4 auth, region should have been resolved, and `region` should not be `None`

aws_mcp_proxy/sigv4_helper.py
wzxxing:
```suggestion
    *, service: str,  region: str, profile: Optional[str] = None,
```

Let's see if we can make region a required argument. create_sigv4_auth expect the region to be resolved lower in the call stack somewhere.

Also adding `*,` because the function take three str arguments, it is easy to mess up during dev.

arangatang: Yeah, ideally Region etc should be done during parse_args (or some new parent function like `resolve_args` that merges cli args with env args and defaults into a dataclass we can pass around)

aws_mcp_proxy/server.py
```diff
@@ -98,12 +101,18 @@ def parse_args():
     parser.add_argument(
         '--profile',
         help='AWS profile to use (uses AWS_PROFILE environment variable if not provided)',
+        default=os.getenv('AWS_PROFILE', 'default'),
+    )
+
+    parser.add_argument(
+        '--region',
+        help='AWS region to use (uses AWS_REGION if not provided)',
```
arangatang, line 110:
```suggestion
        help='AWS region to use (uses AWS_REGION if not provided)',
        default=os.getenv('AWS_REGION', 'default'),
```

aws_mcp_proxy/utils.py
```diff
@@ -23,14 +23,18 @@
 
 
 def create_transport_with_sigv4(
-    url: str, service: str, profile: Optional[str] = None
+    url: str,
+    service: str,
+    profile: Optional[str] = None,
+    region: Optional[str] = None,
```
arangatang: region is not optional at this point.

tests/unit/test_utils.py
```diff
@@ -144,3 +147,45 @@
...
+    """Test cases for determine_aws_region function."""
+
+    def test_determine_region_with_region(self):
+        """Test determination when region is provided."""
+        endpoint = 'https://mcp.us-east-1.api.aws/mcp'
+        region = 'custom-region'
+
+        result = determine_aws_region(endpoint, region)
+
+        assert result == region
```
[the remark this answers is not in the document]
arangatang: Not for this PR, but this check is really noisy, perhaps we can ignore tests for this bandit check.

PR #17, opened 2025-09-30
Remove all mention of unreleased service

kyoncal:
- Remove every mention of unreleased service in docs and code

### DIFF
--- a/aws_mcp_proxy/sigv4_helper.py
+++ b/aws_mcp_proxy/sigv4_helper.py
@@ -186,7 +186,7 @@ def create_sigv4_auth(
 
 
 def create_sigv4_client(
-    service: Optional[str] = None,
+    service: str,
     profile: Optional[str] = None,
     region: Optional[str] = None,
     headers: Optional[Dict[str, str]] = None,
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -206,7 +206,7 @@ class TestServer:
 
         # Act & Assert
         with pytest.raises(ValueError) as exc_info:
-            create_sigv4_client()
+            create_sigv4_client(service='test-service')
         assert 'No AWS credentials found' in str(exc_info.value)
 
     def test_main_module_execution(self):
--- a/tests/test_sigv4_helper.py
+++ b/tests/test_sigv4_helper.py
@@ -304,10 +304,10 @@ class TestCreateSigv4Client:
         mock_client_class.return_value = mock_client
 
         # Test client creation
-        result = create_sigv4_client()
+        result = create_sigv4_client(service='test-service')
 
         # Verify client was created correctly
-        mock_create_auth.assert_called_once_with(None, None, None)
+        mock_create_auth.assert_called_once_with('test-service', None, None)
 
         # Check that AsyncClient was called with correct parameters
         call_args = mock_client_class.call_args
@@ -330,7 +330,7 @@ class TestCreateSigv4Client:
 
         # Test client creation with custom headers
         custom_headers = {'Custom-Header': 'custom-value'}
-        result = create_sigv4_client(headers=custom_headers)
+        result = create_sigv4_client(service='test-service', headers=custom_headers)
 
         # Verify client was created with merged headers
         call_args = mock_client_class.call_args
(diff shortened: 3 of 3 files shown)


aws_mcp_proxy/sigv4_helper.py
```diff
@@ -186,7 +186,7 @@ def create_sigv4_auth(
 
 
 def create_sigv4_client(
-    service: str = 'eks-mcp',
+    service: str = 'your-service',
```
acmlau: I don't think we should do a default service string here

kyoncal: There needs to be some string as the default service. I suggest the empty string.

acmlau: We have https://github.com/aws/aws-mcp-proxy/blob/main/aws_mcp_proxy/utils.py#L54-L83, which ensures we find the correct service string, why can't we leave it without a default? 🤔

kyoncal: The service string will indeed be found and replaced after the fact if none is given, however, given that it is a positional argument, if it's not given a default value, one must be provided. Otherwise uses like [this]( https://github.com/aws/aws-mcp-proxy/blob/main/tests/test_server.py#L209C1-L209C34) won't type check.

acmlau: Rather set it to None then empty string

README.md
```diff
@@ -37,7 +37,7 @@ Optional arguments you can add:
 - `--profile`: AWS profile to use (uses AWS_PROFILE environment variable if not provided)
 - `--allow-write`: Allow tools that require write permissions to be enabled (by default, only tools with the `readOnlyHint` annotation are enabled)
 
-NOTE: `remote-server-url` should be your remote mcp server's URL (including the `/mcp` part). `service-code` should be the service code for your own mcp service, such as `eks-mcp`.
+NOTE: `remote-server-url` should be your remote mcp server's URL (including the `/mcp` part). `service-code` should be the service code for your own mcp service.
```
wzxxing: Could you also change the working in this sentence?

```suggestion
NOTE: `remote-server-url` should be your remote mcp server's URL (including the `/mcp` part). `service-code` should be the service code for the MCP to be connected.
```

aws_mcp_proxy/sigv4_helper.py
```diff
@@ -186,7 +186,7 @@ def create_sigv4_auth(
 
 
 def create_sigv4_client(
-    service: str = 'eks-mcp',
+    service: str = '',
```
wzxxing: Can we make this required, without default parameter? using `""` as the signature code will not work. This will help preventing contributors making mistake

aws_mcp_proxy/server.py
```diff
@@ -74,20 +74,20 @@ def parse_args():
         formatter_class=argparse.RawDescriptionHelpFormatter,
         epilog="""
 Examples:
-  # Run with EKS MCP endpoint
-  aws-mcp-proxy https://eks-mcp.us-west-2.api.aws
+  # Run with your endpoint
+  aws-mcp-proxy <SigV4 MCP endpoint URL>
```
wzxxing, line 78: nit: `AWS Sigv4`

aws_mcp_proxy/sigv4_helper.py
kyoncal: I think the best solution is to make it optional, since it gets inferred if not explicitly provided by the user anyways. In the latest revision, I've made it more elegant by making the string optional as such.
```service: Optional[str] = None,```

aws_mcp_proxy/server.py
kyoncal: There are loads of places in this repo where its referred to as SigV4. Maybe we keep it like this for now >:)

aws_mcp_proxy/sigv4_helper.py
kyoncal: @wzxxing upon further investigation, I believe your solution may be the best. Changing the type from `str` to `Optional[str]` is not possible without major refactor. I agree that making it a required argument reduces confusion (albeit a little less elegant).

cc: @acmlau

aws_mcp_proxy/server.py
arangatang: Think we have to change it to AWS SigV4 at one point, i don't mind now or later.

kyoncal: Maybe a separate PR, since this one has been so drawn out

PR #19, opened 2025-09-30
feat: Add RateLimiting middleware for client calls to proxy and exponential backoff middleware for network mcp.

arangatang:
*Description of changes:*
## Problem
1. We should throttle clients to the proxy who spam it with an unreasonable amount of requests.
2. We should also respect any throttling from the backend servers by doing exponential backoff if they throttle.

## Solution
1. Add `RateLimitingMiddleware` to client facing MCP.
2. Add `RetryMiddleware` to network calling MCP.

Add unit tests (probably an overkill amount)

### DIFF
--- a/tests/unit/test_mcp_proxy_manager.py
+++ b/tests/unit/test_mcp_proxy_manager.py
@@ -404,7 +404,7 @@ class TestRateLimitingBehavior:
 
         call_count = 0
 
-        async def mock_call_next(ctx):
+        async def mock_call_next(context):
             nonlocal call_count
             call_count += 1
             return f'success_{call_count}'
@@ -498,7 +498,7 @@ class TestRetryBehavior:
 
         call_count = 0
 
-        async def failing_then_success(ctx):
+        async def failing_then_success(context):
             nonlocal call_count
             call_count += 1
             if call_count <= 2:  # Fail first 2 attempts
@@ -539,12 +539,14 @@ class TestRetryBehavior:
 
         call_count = 0
 
-        async def timeout_then_success(ctx):
+        async def timeout_then_success(context):
             nonlocal call_count
             call_count += 1
             if call_count == 1:
                 raise TimeoutError('Request timed out')
             return 'success after timeout'
+
+        assert retry_middleware is not None, 'RetryMiddleware should be found'
 
         result = await retry_middleware.on_request(context, timeout_then_success)
 
@@ -578,10 +580,12 @@ class TestRetryBehavior:
 
         call_count = 0
 
-        async def successful_call(ctx):
+        async def successful_call(context):
             nonlocal call_count
             call_count += 1
             return 'success'
+
+        assert retry_middleware is not None, 'RetryMiddleware should be found'
 
         result = await retry_middleware.on_request(context, successful_call)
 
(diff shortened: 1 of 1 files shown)


aws_mcp_proxy/mcp_proxy_manager.py
```diff
@@ -115,3 +120,33 @@ async def _add_prompts(self, proxy: FastMCP) -> None:
...
+        else:
+            self.logger.info('Retry middleware already present')
+
+    def add_rate_limiting_middleware(self, mcp):
+        """Add retry with exponential backoff middleware to target MCP server.
+
+        Args:
+            mcp: The FastMCP instance to add rate limiting to
+        """
+        if not any(isinstance(m, RateLimitingMiddleware) for m in (mcp.middleware or [])):
```
wzxxing: why is this checking required? should this only happen once when the proxy starts?

arangatang: Yeah, just futureproofing it a bit, but probs overkill, i'll remove.

aws_mcp_proxy/mcp_proxy_manager.py
```diff
@@ -115,3 +120,26 @@ async def _add_prompts(self, proxy: FastMCP) -> None:
...
+    def _add_retry_middleware(self, mcp: FastMCP) -> None:
+        """Add retry with exponential backoff middleware to target MCP server.
+
+        Args:
+            mcp: The FastMCP instance to add exponential backoff to
+        """
+        self.logger.info('Adding retry middleware')
+        mcp.add_middleware(RetryMiddleware())
+
+    def add_rate_limiting_middleware(self, mcp):
```
wzxxing: should this be underscore prefixed if it is used only inside of this class?

```suggestion
    def _add_rate_limiting_middleware(self, mcp):
```

aws_mcp_proxy/mcp_proxy_manager.py
```diff
@@ -30,6 +32,7 @@ def __init__(self, target_mcp: FastMCP, read_only: bool = False):
             target_mcp: The target MCP server to add content to
             read_only: If true, disable tools that require write permissions OR that do not have `readOnlyHint` set.
         """
+        self.add_rate_limiting_middleware(target_mcp)
```
wzxxing: same as below


```suggestion
        self._add_rate_limiting_middleware(target_mcp)
```

tests/unit/test_mcp_proxy_manager.py
```diff
@@ -301,3 +305,331 @@
...
+
+        McpProxyManager(mock_target_mcp)
+
+        # Verify rate limiting middleware was added to target MCP
+        mock_target_mcp.add_middleware.assert_called_once()
+
+        # Check that RateLimitingMiddleware was added with correct parameters
+        rate_limiting_call = mock_target_mcp.add_middleware.call_args_list[0]
+        rate_limiting_middleware = rate_limiting_call[0][0]
+        assert isinstance(rate_limiting_middleware, RateLimitingMiddleware)
```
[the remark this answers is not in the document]
wzxxing: We need to figure out a way to disable this bot for unit test

aws_mcp_proxy/mcp_proxy_manager.py
wzxxing, line 39: not relevant to this PR, but I will create a issue to fix this logic. This `add_proxy_content` figure is logically wrong.

aws_mcp_proxy/mcp_proxy_manager.py
arangatang: Yep, moved it around a few times so forgot it, ty!

tests/unit/test_mcp_proxy_manager.py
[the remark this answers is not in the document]
arangatang: Yeah its really annoying. I'll take a look tomorrow.

PR #28, opened 2025-10-03
Fix dynamic tooling

kyoncal:
### Changes

 - Removed `mcp_proxy_manager.py` and redundant MCP server.
 - Moved filtering to middleware
 - Moved middleware adding to `server.py`
 - Fixed and added relevant unit tests
 - Fixed logging by outputting to the correct channel

### User experience

> The user's tools should pass through the proxy transparently instead of being incorrectly cached.
> The user should be able to see proper logs

## Checklist

- These changes were tested manually, alongside the suite of automated tests. The manual tests consisted of creating a simple client to repeated query a mock-server through the proxy to verify that the proxy works and now has the desired behavior of querying the back end instead of caching results.
* [ ] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment

### DIFF
--- a/aws_mcp_proxy/server.py
+++ b/aws_mcp_proxy/server.py
@@ -64,7 +64,9 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
     # Create proxy with the transport
     proxy = FastMCP.as_proxy(transport)
     add_tool_filtering_middleware(proxy, args.read_only)
-    add_retry_middleware(proxy)
+
+    if args.retries:
+        add_retry_middleware(proxy, args.retries)
 
     await proxy.run_async()
 
@@ -84,14 +86,15 @@ def add_tool_filtering_middleware(mcp: FastMCP, read_only: bool = False) -> None
     )
 
 
-def add_retry_middleware(mcp: FastMCP) -> None:
+def add_retry_middleware(mcp: FastMCP, retries: int) -> None:
     """Add retry with exponential backoff middleware to target MCP server.
 
     Args:
         mcp: The FastMCP instance to add exponential backoff to
+        retries: number of retries with which to configure the retry middleware
     """
     logger.info('Adding retry middleware')
-    mcp.add_middleware(RetryMiddleware())
+    mcp.add_middleware(RetryMiddleware(retries))
 
 
 def parse_args():


aws_mcp_proxy/server.py
```diff
@@ -64,12 +66,48 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
...
+    Args:
+        mcp: The FastMCP instance to add rate limiting to
+    """
+    logger.info('Adding rate limiting middleware')
+    mcp.add_middleware(
+        RateLimitingMiddleware(
+            max_requests_per_second=5,
+            burst_capacity=10,
+        )
+    )
```
arangatang: This is going away.

aws_mcp_proxy/middleware/tool_filter.py
```diff
@@ -0,0 +1,43 @@
...
+import logging
+from fastmcp.server.server import FastMCP
+from fastmcp.tools.tool import Tool
+
+class ToolFilteringMiddleware(Middleware): 
+    def __init__(self, read_only: bool, logger: logging.Logger | None = None):
+        self.read_only = read_only 
+        self.logger = logger or logging.getLogger("fastmcp.errors") 
+    
+    async def on_list_tools(self, context: MiddlewareContext, call_next):
```
arangatang: Ccould you add typing to `call_next`

aws_mcp_proxy/middleware/tool_filter.py
```diff
@@ -0,0 +1,43 @@
...
+
+from fastmcp.server.middleware import Middleware, MiddlewareContext
+import logging
+from fastmcp.server.server import FastMCP
+from fastmcp.tools.tool import Tool
+
+class ToolFilteringMiddleware(Middleware): 
+    def __init__(self, read_only: bool, logger: logging.Logger | None = None):
+        self.read_only = read_only 
+        self.logger = logger or logging.getLogger("fastmcp.errors")
```
arangatang: Why not use  `__name__`

aws_mcp_proxy/server.py
arangatang: Rate limiting is going away.

aws_mcp_proxy/middleware/tool_filter.py
kyoncal: Good catch

aws_mcp_proxy/middleware/tool_filter.py
```diff
@@ -0,0 +1,43 @@
...
+            annotations = tool.annotations
+            if self.read_only:
+                # In readOnly mode, skip the tools with no readOnlyHint=True annotation
+                read_only_hint = getattr(annotations, 'readOnlyHint', False)
+                if not read_only_hint:
+                    # Skip tools that don't have readOnlyHint=True
+                    self.logger.info(f'Skipping tool {tool.name} needing write permissions')
+                    continue
+
+            filtered_tools.append(tool)
```
wzxxing, line 55: We can return early by checking 
```python
        if not self.read_only:
            # do not alter the server behaviour if no read only flag
            return tools
```

aws_mcp_proxy/middleware/tool_filter.py
wzxxing: @arangatang ~~can we add this to the typechecker rule somehow?~~

Actually the build failed correctly.

tests/unit/test_tool_filter.py
```diff
@@ -0,0 +1,263 @@
...
+from aws_mcp_proxy.middleware.tool_filter import ToolFilteringMiddleware
+from fastmcp.server.middleware import MiddlewareContext
+
+class MockAnnotationsWithoutReadOnlyHint:
+    """Mock annotations object that raises AttributeError for readOnlyHint."""
+
+    def __getattr__(self, name):
+        if name == 'readOnlyHint':
+            raise AttributeError(f"'{type(self).__name__}' object has no attribute 'readOnlyHint'")
+        return Mock()
```
wzxxing, line 30: Is it possible to simply do the following?

```python
mock = Mock()
mock.readOnlyHint = AttributeError()
```

kyoncal: This didn't work for me. Setting it `readOnlyHint` to an `AttributeError()` didn't cause it to be raised. Instead, it's existence made `readOnlyHint` true and caused to the test to fail because it was supposed to be filtered out. 

```python
    @pytest.mark.asyncio
    async def test_read_only_true_filters_no_hint_tools(
        self, mock_context, read_only_tool, no_hint_tool
    ):
        """Test that read_only=True filters out tools without readOnlyHint."""
        # Arrange
        tools = [read_only_tool, no_hint_tool]
        call_next_mock = AsyncMock(return_value=tools)
        middleware = ToolFilteringMiddleware(read_only=True)
    
        # Act
        result = await middleware.on_list_tools(mock_context, call_next_mock)
    
        # Assert
>       assert len(result) == 1
E       AssertionError: assert 2 == 1
E        +  where 2 = len([<Mock id='4497872240'>, <Mock id='4497871712'>])

tests/unit/test_tool_filter.py:165: AssertionError```

File: aws_mcp_proxy/logging_config.py [hunk 6]
```diff
@@ -34,7 +34,7 @@ def configure_logging(level: Optional[str] = None) -> None:
     date_format = '%Y-%m-%d %H:%M:%S'
 
     # Create console handler with formatting
-    console_handler = logging.StreamHandler(sys.stdout)
+    console_handler = logging.StreamHandler(sys.stderr)
```
wzxxing, line 37: Thank you, good catch!

aws_mcp_proxy/server.py
```diff
@@ -62,10 +63,38 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
...
     proxy = FastMCP.as_proxy(transport)
+    add_tool_filtering_middleware(proxy, args.read_only)
 
-    # Use McpProxyManager to add proxy content
-    proxy_manager = McpProxyManager(local_mcp, args.read_only)
-    await proxy_manager.add_proxy_content(proxy, args.retries)
+    if args.retries:
+        add_retry_middleware(proxy, args.retries)
+
+    await proxy.run_async()
```
wzxxing, line 71: why is this `run_async` lifted to this function?

aws_mcp_proxy/server.py
wzxxing, line 44: `local_mcp` is not used any more, can we directly run this function? Since the fastmcp instance created in `main` is not used anymore?

aws_mcp_proxy/server.py
kyoncal: I directly replaced the first call to `mcp_proxy_manager` with it, although it's original placement outside of this function does make more sense and we can directly run the function.

aws_mcp_proxy/server.py
kyoncal: We'd have to use `asyncio.run` somewhere, since we can't really await the main method, so we'd just be hoisting it elsewhere. I think it makes sense where it is.

aws_mcp_proxy/server.py
```diff
@@ -62,10 +63,38 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
...
+
+
+def add_tool_filtering_middleware(mcp: FastMCP, read_only: bool = False) -> None:
```
arangatang, line 74: Rant for another PR/issue (we should review the code for this in general). 

1. We should not have default values where they are not needed, we should fail rather than accidentally set defaults if omitted. Makes stuff less prone of unexpected behaviour as the code grows. 
1. Ideally we should probably be using kwargs instead of args everywhere, more verbose but much more scalable. 
1. We should use proper typing

Probs I should create an issue for this tbh, when thinking about it.

aws_mcp_proxy/server.py
wzxxing: No, I meant the `local_mcp` is not used any more, we can remove this argument, also remove the FastMCP instance created in main.

wzxxing, line 174: This is not used anymore.

We can rename `setup_mcp_mode` to `run_mcp_proxy`

wzxxing, line 193: Then simply do `asyncio.run(run_mcp_proxy(args))` here

PR #29, opened 2025-10-06
Create Simple MCP Server for Integ tests

JacquesPotgieter:
### Changes

Updates the Integ tests to make them more generic, allowing you to run the test against an MCP Server endpoint, or against AgentCore. This makes it easier to run these integ tests locally, verifying the features still work.

Also created a Simple MCP Server which the tests will be run against. This PR only creates the MCP Server in this package, and as a next step, I'll push the server to agentCore to ensure our Integ test is running against our AWS Account and setup.

### User experience

* No effect to user-experience
* This does make integ testing easier

### Testing

Followed the instructions in `Integ/README.md` to install AgentCore to personal account, and tests succeeded

## Checklist

* [X] I have reviewed the [contributing guidelines](https://github.com/aws/aws-mcp-proxy/blob/main/CONTRIBUTING.md)
* [X] Changes are documented

* [X] Yes - But only to the existing Integ approval workflow
* [ ] No

## Acknowledgment


tests/integ/test_proxy_simple_mcp_server.py
```diff
@@ -0,0 +1,93 @@
...
+    actual_response = await mcp_client.call_tool('elicit_for_my_name', tool_input)
+
+    actual_text = get_text_content(actual_response)
+    failure_message = f"Tool 'elicit_for_my_name' did not return the expected result (Returned {actual_text}) (Expected {expected_response})"
+    assert actual_text == expected_response, failure_message
+
+
+@pytest.mark.asyncio(loop_scope='module')
+async def test_handle_sampling(mcp_client: fastmcp.Client):
+    """TODO."""
```
acmlau, line 92: For when is this to be implemented?

JacqualinPotgieter: Afterwards. The Elicitation and Sampling tests can't currently work with AgentCore. It would only work when running locally. 

Currently, we have an example MCP Server deployed to AgentCore, which does not have any features. I'm planning to replace it with this PR, and then afterwards, we can add the Sampling feature.

tests/integ/README.md
```diff
@@ -0,0 +1,67 @@
+## Integration tests
```
wzxxing, line 1: Thank you!

tests/integ/test_proxy_simple_mcp_server.py
```diff
@@ -0,0 +1,93 @@
...
+    assert actual_text == expected_response, failure_message
+
+
+@pytest.mark.asyncio(loop_scope='module')
+async def test_handle_elicitation_when_accepting(
+    mcp_client: fastmcp.Client, is_using_agentcore: bool
+):
+    """Test calling tool which supports elicitation and accepting it."""
+    if is_using_agentcore:
+        pytest.skip()
```
wzxxing, line 60: why is this skipped? because the agentcore not yet ready? for elicitation?

tests/integ/test_proxy_simple_mcp_server.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.mark.asyncio(loop_scope='module')
+async def test_call_tool(mcp_client: fastmcp.Client):
+    """Test call tool."""
+    name = 'Superman'
+    expected_response = f'Hello {name}'
+
+    tool_input = {'name': name}
+    actual_response = await mcp_client.call_tool('greet', tool_input)
```
wzxxing, line 47: But not every MCP server has a `greet` tool. How would we apply this test on another agentcore ARN or remote endpoint?

tests/integ/test_proxy_simple_mcp_server.py
JacqualinPotgieter: Yes, when I run the elicitation test against a AgentCore hosted version, it hangs. 

So I have to explicitly skip them

tests/integ/test_proxy_simple_mcp_server.py
JacqualinPotgieter: That is why I wanted to add the MCP Server into the repo itself. So the expectation is that people would run these tests against the Simple MCP Server

wzxxing: Ok, I see. I thought we plan to add the test against a real remote MCP server.

JacqualinPotgieter: We might be talking about the same thing here.

Yes, I want people to deploy this MCP Server to test it against a "real" remote MCP Server. But the idea is that we give people a dedicated MCP Server which they can use for this test. Instead of expecting each person to create their own version.

PR #35, opened 2025-10-06
Update readme

acmlau:
## Summary

Updating Readme to align with updated command line arguments and provide overview of how to use proxy server. 

### Changes

* Added different installation methods  
* Added overview of command line arguments
* Explicitly mentioned the use of optional environment variables
* Included the `DEVELOPMENT.md` and `CONTRIBUTING.md`

### User experience

Change of readme still needs to be checked by doc reviewer

## Checklist

* [X] I have reviewed the [contributing guidelines](https://github.com/aws/aws-mcp-proxy/blob/main/CONTRIBUTING.md)
* [ ] Changes are documented

* [ ] Yes
* [X] No

## Acknowledgment

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ The proxy handles SigV4 authentication using local AWS credentials and provides 
 * [Install Python 3.10+](https://www.python.org/downloads/release/python-3100/)
 * [Install the `uv` package manager](https://docs.astral.sh/uv/getting-started/installation/)
 * [Install and configure the AWS CLI with credentials](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html)
-* [Install Docker Desktop](https://www.docker.com/products/docker-desktop) (optional, for Docker usage)
+* (Optional, for docker users) [Install Docker Desktop](https://www.docker.com/products/docker-desktop)
 
 ## Installation
 
@@ -23,7 +23,7 @@ The proxy handles SigV4 authentication using local AWS credentials and provides 
 
 ```
 # Run the server
-aws-mcp-proxy <SigV4 MCP endpoint URL>
+uvx aws-mcp-proxy <SigV4 MCP endpoint URL>
 ```
 
 ### Using Local Repository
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,108 +13,124 @@ The proxy handles SigV4 authentication using local AWS credentials and provides
 * [Install Python 3.10+](https://www.python.org/downloads/release/python-3100/)
 * [Install the `uv` package manager](https://docs.astral.sh/uv/getting-started/installation/)
 * [Install and configure the AWS CLI with credentials](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html)
+* [Install Docker Desktop](https://www.docker.com/products/docker-desktop) (optional, for Docker usage)
```
JacqualinPotgieter: Should we add the optional in the beginning? To avoid confusion, or people skim reading.

README.md
```diff
@@ -15,108 +13,124 @@ The proxy handles SigV4 authentication using local AWS credentials and provides
...
+
+## Configuration Parameters
 
-#### Details
+|Parameter	|Description	|Default	|Required	|
+|---	|---	|---	|---	|
+|`endpoint`	|MCP endpoint URL (e.g., `https://your-service.us-east-1.amazonaws.com/mcp`)	|N/A	|Yes	|
+|---	|---	|---	|---	|
+|`--service`	|AWS service name for SigV4 signing	|Inferred from endpoint if not provided	|No	|
+|`--profile`	|AWS profile for AWS credentials to use	|Uses `AWS_PROFILE` environment variable if not set, defaults to `default`	|No	|
```
JacqualinPotgieter: I updated this yesterday, as there was a bug. We currently fallback to the env param, and that falls back to default Botocore behaviour. See https://github.com/aws/aws-mcp-proxy/pull/33

I updated the wording to only say: "AWS profile to use (uses AWS_PROFILE environment variable if not provided)"

README.md
```diff
@@ -15,108 +13,124 @@ The proxy handles SigV4 authentication using local AWS credentials and provides
...
+|`--region`	|AWS region to use	|Uses `AWS_REGION` environment variable if not set, defaults to `us-east-1`	|No	|
+|`--read-only`	|Disable tools which may require write permissions (tools which DO NOT require write permissions are annotated with [`readOnlyHint=true`](https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations-readonlyhint))|`True`	|No	|
```
JacqualinPotgieter: By default this is False. In the default case, we want people to be able to use all of their tools.

README.md
```diff
@@ -15,108 +13,124 @@ The proxy handles SigV4 authentication using local AWS credentials and provides
...
 
-MCP mode provides a streamlined connection to a single backend MCP server.
+## Optional Environment Variables
 
-#### 1. Configure MCP Client
+Set the environment variables for the AWS MCP Proxy:
 
-Add this to your MCP client configuration, replacing env variables to match the AWS credentials and region you want to use:
+```
+export AWS_PROFILE=<YOUR AWS PROFILE>
```
JacqualinPotgieter: Do we need to mention the other cred env variables? 

Since customers can use either `AWS_PROFILE` or the `AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN`

For example, see how `aws-api-mcp-server` does it. https://github.com/awslabs/mcp/tree/main/src/aws-api-mcp-server#%EF%B8%8F-configuration-options

README.md
acmlau:
```
    parser.add_argument(
        '--read-only',
        action='store_true',
        help='Disable tools which may require write permissions (readOnlyHint True or unknown)',
    )
```

But we store true as a value?

wzxxing: by default this is `None` (falsy), when `--read-only` is in the args, this becomes `True`.

README.md
```diff
@@ -15,108 +13,124 @@ The proxy handles SigV4 authentication using local AWS credentials and provides
...
+## Installation
+
+### Using PyPi
 
-## Getting Started
+*Note: will work after publishing to PyPi*
+
+```
+# Run the server
+aws-mcp-proxy <SigV4 MCP endpoint URL>
```
wzxxing: We want the users to run the package using `uvx` so that they always use the latest version.

README.md
```diff
@@ -15,108 +13,124 @@ The proxy handles SigV4 authentication using local AWS credentials and provides
...
-- `--retries`: Configures number of retries done when calling upstream services, setting this to 0 disables retries.
+## Setup Examples
 
-NOTE: `remote-server-url` should be your remote mcp server's URL (including the `/mcp` part). `service-code` should be the service code for the MCP to be connected.
+Add the following configuration to your MCP client config file (e.g., for Amazon Q Developer CLI, edit `~/.aws/amazonq/mcp.json`):
+**Note** Add your own endpoint by replacing  `<SigV4 MCP endpoint URL>`
 
-Example with all options
-```json
+### Running from local - using uv
```
wzxxing, line 79: Do we plan to add the option of running the package using `uvx` after publishing to public?

acmlau: We can already add it, but then we need to add another disclaimer like ""Will work after publishing to PyPi"

README.md
acmlau: Thank you, will update

PR #37, opened 2025-10-07
fix(server.py,-utils.py): fix issue where regions weren't parsed prop…

kyoncal:
## Summary

Ensure that regions are resolved in the following order: 
1. from --region param 
2. from url 
3. from environment variable and throws an error if none of these are available.

### Changes

> the logic in `determine_aws_regions` was changed to reflect desired behavior.

### User experience

> The regions should be resolved in a more logical order and fail if no region can be resolved.

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment

### DIFF
--- a/aws_mcp_proxy/utils.py
+++ b/aws_mcp_proxy/utils.py
@@ -15,11 +15,16 @@
 """Utility functions for the AWS MCP Proxy."""
 
 import httpx
+import logging
+import os
 import re
 from aws_mcp_proxy.sigv4_helper import create_sigv4_client
 from fastmcp.client.transports import StreamableHttpTransport
 from typing import Dict, Optional
 from urllib.parse import urlparse
+
+
+logger = logging.getLogger(__name__)
 
 
 def create_transport_with_sigv4(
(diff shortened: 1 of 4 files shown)


aws_mcp_proxy/utils.py
```diff
@@ -114,11 +120,16 @@ def determine_aws_region(endpoint: str, region: Optional[str] = None) -> str:
...
+    if region_match:
+        logger.debug('Region determined through endpoint URL')
+        return region_match.group(1)
+
+    environment_region = os.getenv('AWS_REGION')
+    if environment_region:
+        logger.debug('Region determined through environment variable')
+        return environment_region
+
+    raise ValueError(
```
wzxxing, line 132: We should also consider the global endpoints might use `us-east-1` or [sigv4a](https://github.com/aws-samples/sigv4a-signing-examples/blob/main/python/sigv4a_sign.py) as the signature. 

But this is good enough for the bugfix! Thanks

PR #44, opened 2025-10-08
test: add integration test for dynamic tooling

kyoncal:
### Changes

> Add test to ensure that the proxy does not cache list_tool states

### User experience

> Add integration test

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment

### DIFF
--- a/tests/integ/test_proxy_dynamic_tools.py
+++ b/tests/integ/test_proxy_dynamic_tools.py
@@ -9,8 +9,11 @@ logger = logging.getLogger(__name__)
 
 
 @pytest.mark.asyncio(loop_scope='module')
-async def test_proxy_reflects_tool_addition(mcp_client: fastmcp.Client):
+async def test_proxy_reflects_tool_addition(mcp_client: fastmcp.Client, is_using_agentcore: bool):
     """Test that when backend dynamically adds a tool, proxy reflects it on next list_tools call."""
+    if is_using_agentcore:
+        pytest.skip()
+
     # Arrange - Get initial tool list
     initial_tools = await mcp_client.list_tools()
     initial_tool_names = [tool.name for tool in initial_tools]


tests/integ/mcp/simple_mcp_server/mcp_server.py
```diff
@@ -65,7 +87,7 @@ def main():
     mcp.run(
         transport='http',
         host='0.0.0.0',
-        port=8000,
+        port=8002,
```
JacqualinPotgieter: What issue did you face that you need to change this?

Unfortunately, we must have this set to 8000 when hosting with AgentCore.

tests/integ/mcp/simple_mcp_server/mcp_server.py
```diff
@@ -20,6 +20,28 @@ def greet(name: str):
     return f'Hello {name}'
 
 
+##### Dynamic Tool Testing
```
JacqualinPotgieter, line 23: Nice work here. Once you merge your change, we will need to update the deployed version as well. I can help with that though

tests/integ/mcp/simple_mcp_server/mcp_server.py
```diff
@@ -20,6 +20,28 @@ def greet(name: str):
...
+
+_multiply_registered = False
+
+
+@mcp.tool
+def add_tool_multiply():
+    """MCP Tool used for testing dynamic tool behavior through the proxy."""
+    global _multiply_registered
+
+    if not _multiply_registered:
```
wzxxing: This probably works fine for testing our own MCP server. However, in a real MCP server, after the tool list in the server changes, the server should send a notification via 
[notifications/tools/list_changed](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#list-changed-notification). The `ctx` parameter in the fastmcp functional tools has a method to send this notification.

There are issue with bedrock agentcore when hosting a MCP server capable of doing this so this is not a blocking comment, just so that you are aware of it.

JacqualinPotgieter: If this won't work in AgentCore, we should skip this test when running against AgentCore. Similar to the Elicitation test.

tests/integ/mcp/simple_mcp_server/mcp_server.py
kyoncal: This was a silly mistake on my end and should not have been committed. Fixed in latest revision.

tests/integ/mcp/simple_mcp_server/mcp_server.py
wzxxing: This will work. The notification is the server asking the client to call `tools/list` again, but in our test, we always list again, so this is not a problem for our tests.

tests/integ/mcp/simple_mcp_server/mcp_server.py
```diff
@@ -20,6 +20,28 @@ def greet(name: str):
     return f'Hello {name}'
 
 
+##### Dynamic Tool Testing
+
+_multiply_registered = False
```
wzxxing: Instead of relying on a global variable, you can use https://gofastmcp.com/servers/context#state-management.

tests/integ/mcp/simple_mcp_server/mcp_server.py
```diff
@@ -65,7 +83,7 @@ def main():
     mcp.run(
         transport='http',
         host='0.0.0.0',
-        port=8000,
+        port=8015,
```
wzxxing: unfortunately the bedrock runtime only understands the mcp server if it runs on 8000.

> When you configure a Amazon Bedrock AgentCore Runtime with the MCP protocol, the service expects MCP server containers to be available at the path 0.0.0.0:8000/mcp, which is the default path supported by most official MCP server SDKs.

https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-mcp.html



```suggestion
        port=8000,
```

tests/integ/mcp/simple_mcp_server/mcp_server.py
```diff
@@ -20,6 +20,24 @@ def greet(name: str):
     return f'Hello {name}'
 
 
+##### Dynamic Tool Testing
+
+
+@mcp.tool
+def add_tool_multiply(ctx: Context):
+    """MCP Tool used for testing dynamic tool behavior through the proxy."""
+    if not ctx.get_state('_multiply_registered'):
```
wzxxing: nitpicking, the underscore prefix is not needed anymore.,

tests/integ/test_proxy_dynamic_tools.py
```diff
@@ -0,0 +1,51 @@
...
+import logging
+import pytest
+
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.asyncio(loop_scope='module')
+async def test_proxy_reflects_tool_addition(mcp_client: fastmcp.Client, is_using_agentcore: bool):
+    """Test calling tool which supports elicitation and accepting it."""
```
JacqualinPotgieter: I think you copied over the Elicitation comment :-)

kyoncal: Yes... Apologies for the silly mistake.

tests/integ/mcp/simple_mcp_server/mcp_server.py
kyoncal: Fixed!

tests/integ/mcp/simple_mcp_server/mcp_server.py
kyoncal: This was, again, a silly mistake on my part.. Thanks!

tests/integ/mcp/simple_mcp_server/mcp_server.py
kyoncal: FWIW, the test is being skipped regardless because of AgentCore does not support state, so we cannot test the dynamic addition.

PR #62, opened 2025-10-30
feat(sigv4_helper): inject AWS_REGION in _meta

kyoncal:
## Summary

Implemented automatic AWS_REGION metadata injection into MCP JSON-RPC requests using `httpx` request hooks. Added `_inject_metadata_hook` function to sigv4_helper.py that injects metadata into the _meta field of request params, merging with existing metadata when present (existing values take precedence). Includes unit tests and an end-to-end integration test with a new echo_metadata tool in the test MCP server to verify metadata transmission.

### User experience

AWS_REGION is now automatically injected into all MCP requests transparently - no manual configuration required. The proxy adds region information to the _meta field, making it available to MCP servers without any client-side code changes.

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

- [x] Did integration tests succeed? (Only locally, ACR doesn't work for me on this commit or main head)

## Acknowledgment

### DIFF
--- a/tests/integ/test_proxy_simple_mcp_server.py
+++ b/tests/integ/test_proxy_simple_mcp_server.py
@@ -124,3 +124,7 @@ async def test_metadata_injection_aws_region(
     assert 'AWS_REGION' in response_data['received_meta'], (
         f'Metadata should contain AWS_REGION: {response_data["received_meta"]}'
     )
+    assert (
+        response_data['received_meta']['AWS_REGION']
+        == remote_mcp_server_configuration['region_name']
+    ), f'AWS_REGION should be {remote_mcp_server_configuration["region_name"]}'
--- a/mcp_proxy_for_aws/server.py
+++ b/mcp_proxy_for_aws/server.py
@@ -55,13 +55,20 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
 
     # Validate and determine region
     region = determine_aws_region(args.endpoint, args.region)
+    forwarding_region = args.forwarding_region or region
     logger.debug('Using region: %s', region)
 
     # Get profile
     profile = args.profile
 
     # Log server configuration
-    logger.info('Using service: %s, region: %s, profile: %s', service, region, profile)
+    logger.info(
+        'Using service: %s, region: %s, forwarding region: %s, profile: %s',
+        service,
+        region,
+        forwarding_region,
+        profile,
+    )
     logger.info('Running in MCP mode')
 
     timeout = httpx.Timeout(
@@ -72,7 +79,9 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
     )
 
     # Create transport with SigV4 authentication
-    transport = create_transport_with_sigv4(args.endpoint, service, region, timeout, profile)
+    transport = create_transport_with_sigv4(
+        args.endpoint, service, region, forwarding_region, timeout, profile
+    )
 
     # Create proxy with the transport
     proxy = FastMCP.as_proxy(transport)
@@ -163,7 +172,13 @@ Examples:
 
     parser.add_argument(
         '--region',
-        help='AWS region to use (uses AWS_REGION environment variable if not provided, with final fallback to us-east-1)',
+        help='AWS region to sign (uses AWS_REGION environment variable if not provided, with final fallback to us-east-1)',
+        default=None,
+    )
+
+    parser.add_argument(
+        '--forwarding-region',
+        help='AWS region to forward to server (uses --region if not provided)',
         default=None,
     )
 
--- a/mcp_proxy_for_aws/utils.py
+++ b/mcp_proxy_for_aws/utils.py
@@ -32,6 +32,7 @@ def create_transport_with_sigv4(
     url: str,
     service: str,
     region: str,
+    forwarding_region: str,
     custom_timeout: httpx.Timeout,
     profile: Optional[str] = None,
 ) -> StreamableHttpTransport:
@@ -41,6 +42,7 @@ def create_transport_with_sigv4(
         url: The endpoint URL
         service: AWS service name for SigV4 signing
         region: AWS region to use
+        forwarding_region: AWS region to forward to server
         custom_timeout: httpx.Timeout used to connect to the endpoint
         profile: AWS profile to use (optional)
 
@@ -60,7 +62,7 @@ def create_transport_with_sigv4(
             region=region,
             headers=headers,
             timeout=custom_timeout,
-            metadata={'AWS_REGION': region},
+            metadata={'AWS_REGION': forwarding_region},
             auth=auth,
         )
 
--- a/tests/integ/conftest.py
+++ b/tests/integ/conftest.py
@@ -77,7 +77,7 @@ def _build_endpoint_environment_remote_configuration():
 
     region_name = os.environ.get('AWS_REGION')
     if not region_name:
-        logger.warn('AWS_REGION param not set. Defaulting to us-east-1')
+        logger.warning('AWS_REGION param not set. Defaulting to us-east-1')
         region_name = 'us-east-1'
 
     logger.info(f'Starting server with config - {remote_endpoint_url=} and {region_name=}')
--- a/tests/unit/test_server.py
+++ b/tests/unit/test_server.py
@@ -56,6 +56,7 @@ class TestServer:
         mock_args.profile = None
         mock_args.read_only = True
         mock_args.retries = 1
+        mock_args.forwarding_region = None
         # Add timeout parameters
         mock_args.timeout = 180.0
         mock_args.connect_timeout = 60.0
@@ -86,8 +87,9 @@ class TestServer:
         assert call_args[0][0] == 'https://test.example.com'
         assert call_args[0][1] == 'test-service'
         assert call_args[0][2] == 'us-east-1'
-        # call_args[0][3] is the Timeout object
-        assert call_args[0][4] is None  # profile
+        assert call_args[0][3] == 'us-east-1'  # forwarding_region (defaults to region)
+        # call_args[0][4] is the Timeout object
+        assert call_args[0][5] is None  # profile
         mock_as_proxy.assert_called_once_with(mock_transport)
         mock_add_filtering.assert_called_once_with(mock_proxy, True)
         mock_add_retry.assert_called_once_with(mock_proxy, 1)
@@ -116,6 +118,7 @@ class TestServer:
         mock_args.profile = 'test-profile'
         mock_args.read_only = False
         mock_args.retries = 0  # No retries
+        mock_args.forwarding_region = 'eu-west-1'
         # Add timeout parameters
         mock_args.timeout = 180.0
         mock_args.connect_timeout = 60.0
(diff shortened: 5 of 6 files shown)


mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +122,59 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+    if request.content and metadata:
+        try:
+            # Parse the request body
+            body = json.loads(await request.aread())
+
+            # Check if it's a JSON-RPC request
+            if isinstance(body, dict) and 'jsonrpc' in body:
+                # Ensure params exists
+                # if 'params' not in body:
+                #     body['params'] = {}
```
arangatang: Should this be commented out?

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +122,59 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+                # Merge metadata (existing takes precedence)
+                if isinstance(existing_meta, dict):
+                    body['params']['_meta'] = {**metadata, **existing_meta}
+                else:
+                    body['params']['_meta'] = metadata
+
+                # Create new content
+                new_content = json.dumps(body).encode('utf-8')
+
+                from httpx._content import ByteStream
```
arangatang: Lets not import in the middle of a function.

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +122,59 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+
+                request.stream = ByteStream(new_content)
+                request.headers['Content-Length'] = str(len(new_content))
+
+                logger.info('Injected metadata into _meta: %s', body['params']['_meta'])
+
+        except (json.JSONDecodeError, KeyError, TypeError) as e:
+            # Not a JSON request or invalid format, skip metadata injection
+            logger.debug('Skipping metadata injection: %s', e)
```
arangatang: Lets make this `error` level.

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +122,59 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+            if isinstance(body, dict) and 'jsonrpc' in body:
+                # Ensure params exists
+                # if 'params' not in body:
+                #     body['params'] = {}
+
+                # Ensure _meta exists in params
+                if '_meta' not in body['params']:
+                    body['params']['_meta'] = {}
+                else:
+                    print()
```
arangatang: Why this print?

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +122,59 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+
+                # Get existing metadata
+                existing_meta = body['params']['_meta']
+
+                # Merge metadata (existing takes precedence)
+                if isinstance(existing_meta, dict):
+                    body['params']['_meta'] = {**metadata, **existing_meta}
+                else:
+                    body['params']['_meta'] = metadata
```
arangatang, line 227: nit: I think this can be replaced with
```
body['params']['_meta']  = metadata | body['params']['_meta'] 
```

tests/unit/test_sigv4_metadata_injection.py
```diff
@@ -0,0 +1,226 @@
...
+            {'jsonrpc': '2.0', 'id': 1, 'method': 'tools/call', 'params': {'name': 'myTool'}}
+        ).encode('utf-8')
+
+        request = httpx.Request('POST', 'https://example.com/mcp', content=request_body)
+
+        # Call the hook
+        await _inject_metadata_hook(metadata, request)
+
+        # Clear cached content and read from the new stream
+        del request._content
```
arangatang: Why the delete here?

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +122,59 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+                    body['params']['_meta'] = {}
+                else:
+                    print()
+
+                # Get existing metadata
+                existing_meta = body['params']['_meta']
+
+                # Merge metadata (existing takes precedence)
+                if isinstance(existing_meta, dict):
+                    body['params']['_meta'] = {**metadata, **existing_meta}
```
arangatang, line 224: I think we may need to log if there are failed overwrites here, i.e. if a key was available in both, then this would make it clearer what happened.

tests/integ/test_proxy_simple_mcp_server.py
```diff
@@ -91,3 +92,35 @@ async def test_handle_elicitation_when_declining(
 async def test_handle_sampling(mcp_client: fastmcp.Client):
     """TODO."""
     pass
+
+
+@pytest.mark.asyncio(loop_scope='module')
+async def test_metadata_injection_aws_region(
```
arangatang, line 98: Thanks for the integ test!

tests/unit/test_sigv4_metadata_injection.py
```diff
@@ -0,0 +1,226 @@
...
+        stream_content = await request.aread()
+
+        modified_body = json.loads(stream_content.decode('utf-8'))
+
+        # Existing metadata takes precedence
+        assert modified_body['params']['_meta']['field1'] == 'existing'
+        assert modified_body['params']['_meta']['field2'] == 'original'
+        assert modified_body['params']['_meta']['AWS_REGION'] == 'us-west-2'
+
+    @pytest.mark.asyncio
```
arangatang: Duplicate decorator.

mcp_proxy_for_aws/sigv4_helper.py
kyoncal: I believe so. There is no MCP request that shouldn't contains a params field.

mcp_proxy_for_aws/sigv4_helper.py
kyoncal: Agreed. Will fix.

mcp_proxy_for_aws/sigv4_helper.py
kyoncal: oops. remnant from debugging.

tests/unit/test_sigv4_metadata_injection.py
kyoncal: ~After looking at the implementation of `request.aread()`, I've reached the conclusion that it is indeed unnecessary.~ I've moved it to `sigv4_helper.py`, because it is necessary. Since `_content` functions as a cache, it seems asthough the cached value is read instead of the updated one. Although it doesn't make sense for the caller to need to do this.

mcp_proxy_for_aws/sigv4_helper.py
arangatang: Then either delete it, or uncomment it. I'd leave it there, fix it and maybe log a warning if this is expected?

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
             raise e
 
 
+def _resign_request_with_sigv4(
```
arangatang, line 126: Tbh, if we go with this approach we should not sign it at all before this point and only do signing here. I.e. no need to re-sign.

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+    """
+    # Remove old signature headers before re-signing
+    headers_to_remove = ['Content-Length', 'x-amz-date', 'x-amz-security-token', 'authorization']
+    for header in headers_to_remove:
+        request.headers.pop(header, None)
+
+    # Set the new Content-Length
+    request.headers['Content-Length'] = str(len(request.content))
+
+    logger.info('Headers after cleanup: %s', request.headers)
```
arangatang, line 151: Don't think we should log this anymore, atleast not with info level.

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
             raise e
 
 
+def _resign_request_with_sigv4(
+    request: httpx.Request,
+    region: str,
+    service: str,
+    profile: Optional[str] = None,
+) -> None:
+    """Re-sign an HTTP request with AWS SigV4 after content modification.
```
arangatang, line 132: We should probably only do signing here and not re-sign. But this changes the core way the proxy works. We need to align on this.

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+
+    # Set the new Content-Length
+    request.headers['Content-Length'] = str(len(request.content))
+
+    logger.info('Headers after cleanup: %s', request.headers)
+
+    # Get AWS credentials
+    session = create_aws_session(profile)
+    credentials = session.get_credentials()
+    logger.info('Re-signing request with credentials for access key: %s', credentials.access_key)
```
arangatang, line 156: This should not be logged / be on debug level.

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+    signer = SigV4Auth(credentials, service, region)
+    aws_request = AWSRequest(
+        method=request.method,
+        url=str(request.url),
+        data=request.content,
+        headers=headers_for_signing,
+    )
+
+    # Sign the request
+    logger.info('AWS request before signing: %s', aws_request.headers)
```
arangatang, line 172: Same as above, should be `debug`

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+    signer.add_auth(aws_request)
+    logger.info('AWS request after signing: %s', aws_request.headers)
```
arangatang, line 174: Same as above, should be `debug`

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+
+    # Update request headers with signed headers
+    request.headers.update(dict(aws_request.headers))
+    logger.info('Request headers after re-signing: %s', request.headers)
```
arangatang, line 178: Same as above, should be `debug`

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+                            logger.warning(
+                                'Metadata key "%s" already exists in _meta. '
+                                'Keeping existing value "%s", ignoring injected value "%s"',
+                                key,
+                                existing_meta[key],
+                                metadata[key],
+                            )
+                    body['params']['_meta'] = {**metadata, **existing_meta}
+                else:
+                    logger.info('Replacing non-dict _meta value with injected metadata')
```
arangatang, line 226: What do you mean with `non-dict` here? 
Maybe: 
```
'No conclicting _meta entries exist.`
```

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+                new_content = json.dumps(body).encode('utf-8')
+
+                # Update the request with new content
+                request.stream = ByteStream(new_content)
+                request._content = new_content
+
+                # Re-sign the request with the new content
+                _resign_request_with_sigv4(request, region, service)
+
+                logger.info('Injected metadata into _meta: %s', body['params']['_meta'])
```
arangatang, line 239: nit: Maybe`Updated _meta after injection of additional _meta parameters: <meta>`

PR #68, opened 2025-11-01
fix: correct log_level type annotation  in add_logging_middleware

mazyu36:
## Summary

Fix type annotation mismatch in `add_logging_middleware` function. 

The `log_level` parameter was incorrectly annotated as `int` but actually receives string values (`'DEBUG'`, `'INFO'`, `'WARNING'`, `'ERROR'`, `'CRITICAL'`) from argparse.

### Changes

Changed `log_level` parameter type annotation from `int` to `str` in `add_logging_middleware()` function in `mcp_proxy_for_aws/server.py`

### User experience

No user-facing changes. 

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/server.py
+++ b/mcp_proxy_for_aws/server.py
@@ -111,7 +111,7 @@ def add_retry_middleware(mcp: FastMCP, retries: int) -> None:
     mcp.add_middleware(RetryMiddleware(retries))
 
 
-def add_logging_middleware(mcp: FastMCP, log_level: int) -> None:
+def add_logging_middleware(mcp: FastMCP, log_level: str) -> None:
     """Add logging middleware."""
     if log_level != 'DEBUG':
         return


mcp_proxy_for_aws/server.py
```diff
@@ -111,7 +111,7 @@ def add_retry_middleware(mcp: FastMCP, retries: int) -> None:
     mcp.add_middleware(RetryMiddleware(retries))
 
 
-def add_logging_middleware(mcp: FastMCP, log_level: int) -> None:
+def add_logging_middleware(mcp: FastMCP, log_level: str) -> None:
```
wzxxing, line 114: Thanks for your PR!

I think the reason why this was `int` because logging levels such as `logging.DEBUG` is defined as a integer. but this function should accept a string because the level is from the command line argument.

PR #71, opened 2025-11-04
Forward region via meta

kyoncal:
Extension of #62 
## Summary

Implemented automatic AWS_REGION metadata injection into MCP JSON-RPC requests using `httpx` request hooks. Added `_inject_metadata_hook` function to `hooks.py` that injects metadata into the _meta field of request params, merging with existing metadata when present (existing values take precedence). Added optional metadata argument. Includes unit tests and an end-to-end integration test with a new echo_metadata tool in the test MCP server to verify metadata transmission.

### User experience

AWS_REGION is now automatically injected into all MCP requests transparently - no manual configuration required. The proxy adds region information to the _meta field, making it available to MCP servers without any client-side code changes.

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment


mcp_proxy_for_aws/cli.py
```diff
@@ -60,7 +60,13 @@ def parse_args():
...
-        help='AWS region to use (uses AWS_REGION environment variable if not provided, with final fallback to us-east-1)',
+        help='AWS region to sign (uses AWS_REGION environment variable if not provided, with final fallback to us-east-1)',
+        default=None,
+    )
+
+    parser.add_argument(
+        '--forwarding-region',
+        help='AWS region to forward to server (uses --region if not provided)',
         default=None,
     )
```
wzxxing, line 110: These options would make the config more restricted. I suggest that we use an aggregated type here such as a dictionary or json string

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,6 +123,126 @@ async def _handle_error_response(response: httpx.Response) -> None:
...
+    logger.info('AWS request before signing: %s', aws_request.headers)
+    signer.add_auth(aws_request)
+    logger.info('AWS request after signing: %s', aws_request.headers)
+
+    # Update request headers with signed headers
+    request.headers.update(dict(aws_request.headers))
+    logger.info('Request headers after re-signing: %s', request.headers)
+
+
+async def _inject_metadata_hook(
```
wzxxing: Let's move these code into a dedicated http module.

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -228,5 +353,8 @@ def create_sigv4_client(
     return httpx.AsyncClient(
         auth=sigv4_auth,
         **client_kwargs,
-        event_hooks={'response': [_handle_error_response]},
+        event_hooks={
+            'response': [_handle_error_response],
+            'request': [partial(_inject_metadata_hook, metadata or {}, region, service)],
```
wzxxing: per the mcp spec, the `_meta` field is https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta


Should we communicate with the teams owning remote MCP that we will use their endpoint as the key in meta? The resulted meta field would look like:

```json
{
    "params": {
        "_meta": {
            "some-mcp-server.us-east-1.amazonaws.com/config": {
                "key": "value",
                "from": "cli-args"
            }
        }
    }
}
```

kyoncal: Using their endpoint as a key could be an idea for the future, however, I don't see it adding much value right, especially since there is currently only one metadata field that we set.

wzxxing: Optionally, our cli program can take an argument `--config-key` as the key in `_meta`.

wzxxing: Yes, the prefix is optional. But then the config key needs to be chosen carefully.

kyoncal: To my knowledge, the only thing that matters with regard to naming is that these rules are followed and that the user doesn't break them:

> Prefix:
> 
>     If specified, MUST be a series of labels separated by dots (.), followed by a slash (/).
>         Labels MUST start with a letter and end with a letter or digit; interior characters can be letters, digits, or hyphens (-).
>     Any prefix beginning with zero or more valid labels, followed by modelcontextprotocol or mcp, followed by any valid label, is reserved for MCP use.
>         For example: modelcontextprotocol.io/, mcp.dev/, api.modelcontextprotocol.org/, and tools.mcp.com/ are all reserved.
> 
> Name:
> 
>     Unless empty, MUST begin and end with an alphanumeric character ([a-z0-9A-Z]).
>     MAY contain hyphens (-), underscores (_), dots (.), and alphanumerics in between.

Perhaps we can add some sort of sanitation? Changing the spec that other people expect at this point might not be optimal

mcp_proxy_for_aws/server.py
```diff
@@ -54,11 +54,22 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
     region = determine_aws_region(args.endpoint, args.region)
     logger.debug('Using region: %s', region)
 
+    # Build metadata dictionary - start with AWS_REGION, then merge user metadata
+    metadata = {'AWS_REGION': region}
```
JacqualinPotgieter, line 58: Is there a reason why we are always setting this metadata? 

I'm just wondering if we shouldn't set nothing in the default case, and users must specify the param if they want to use this

mcp_proxy_for_aws/server.py
```diff
@@ -54,11 +54,22 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
...
     # Get profile
     profile = args.profile
 
     # Log server configuration
-    logger.info('Using service: %s, region: %s, profile: %s', service, region, profile)
+    logger.info(
+        'Using service: %s, region: %s, forwarding region: %s, profile: %s',
+        service,
+        region,
+        metadata.get('AWS_REGION'),
```
JacqualinPotgieter: Should we log the metadata itself? Right now, we want to send AWS_REGION, but there might be other params in the future.

mcp_proxy_for_aws/server.py
kyoncal: From my understanding of the intended workflow, there is no variation wherein an `AWS_REGION` is not necessary. If this is not the case, I'm happy to make it entirely optional.

JacqualinPotgieter: Let's maybe differentiate between "feature-request" and "how other projects will use it". The "feature-request" here is to add the ability to add metadata. 

With that in mind, I think let's make it completely optional

tests/unit/test_hooks.py
```diff
@@ -0,0 +1,421 @@
...
+    mock_session = MagicMock()
+    mock_credentials = MagicMock()
+    mock_credentials.access_key = 'test-access-key'
+    mock_credentials.secret_key = 'test-secret-key'
+    mock_credentials.token = 'test-token'
+    mock_session.get_credentials.return_value = mock_credentials
+    return mock_session
+
+
+class TestHandleErrorResponse:
```
arangatang, line 55: nit: While you are moving these tests, can you fix them? Currently these tests don't actually check anything.

mcp_proxy_for_aws/hooks.py
```diff
@@ -0,0 +1,171 @@
...
+    This should be the last hook called to ensure the signature includes any modifications.
+
+    Args:
+        region: AWS region for SigV4 signing
+        service: AWS service name for SigV4 signing
+        profile: AWS profile to use (optional)
+        request: The HTTP request object to sign (modified in-place)
+    """
+    # Import here to avoid circular dependency and for compatibility
+    from mcp_proxy_for_aws.sigv4_helper import SigV4HTTPXAuth, create_aws_session
```
arangatang: pls fix: Why do we import within a hook? Is there no way around this? Does this mean that for every API call we will be re-importing this.

kyoncal: Outside of this, the only other fix I had found was moving everything to the same module, although I'm not sure that would be better?

arangatang (approved):
I did like the hooks seperation, maybe we can reshuffle again later without the circular dep. For now this looks ok. Thanks for the work @kyoncal!

JacquesPotgieter (approved):

PR #74, opened 2025-11-05
feat(config): add yaml config file support

konippi:
### Changes

Add support for loading configuration from YAML files, allowing users to manage complex configurations more easily. CLI arguments take precedence over file configuration for maximum flexibility.

### User experience

#### Before

```bash
# Long command line with many arguments
mcp-proxy-for-aws https://example.com/mcp \
  --service my-service \
  --profile production \
  --region us-east-1 \
  --read-only \
  --log-level DEBUG \
  --retries 3 \
  --timeout 200
```

#### After

```bash
# Simple command with config file
mcp-proxy-for-aws --config config.yaml
```

## Checklist

* [x] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/cli.py
+++ b/mcp_proxy_for_aws/cli.py
@@ -82,8 +82,8 @@ Examples:
     parser.add_argument(
         '--log-level',
         choices=['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'],
-        default='INFO',
-        help='Set the logging level (default: INFO)',
+        default='ERROR',
+        help='Set the logging level (default: ERROR)',
     )
 
     parser.add_argument(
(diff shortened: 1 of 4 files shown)


wzxxing:
Hi @Konippi, thanks a lot for your PR.

Our project was only recently released so we did not have a mechanism for RFCs. This is a new feature and ideally we would like you to open a issue in the project. I will add this to the contributing guideline soon.

We do have plan to introduce a config feature, but it is not for local config, but the server-side config. And we are thinking of using `--config` parameter as a "append` argument (can be specified many times). e.g.

```
mcp-proxy-for-aws --config AWS_REGION=$AWS_REGION \
    --config readonly=true \
    --config timeout=30s
    <endpoint>
```

The config will be added to the JSON RPC message to send to the remote server.

PR #79, opened 2025-11-06
fix(mcp_proxy_for_aws/utils.py): add override for bedrock-agentcore service name detection

arangatang:
Fixes failures when connecting to an Agentcore service.

## Summary
Bedrock AgentCore url structure does not align to our current `--service` detection regex. For now i'm adding an override if we realize we will need more overrides as time goes on we should re-evaluate the regex itself.

### Changes

Adding a check for `.bedrock-agentcore.` in the url and using this as service if found. 

### User experience

After fix users connecting to AgentCore can omit the `--service` flag.

## Checklist

* [X] Changes are documented

* [ ] Yes
* [X] No

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/utils.py
+++ b/mcp_proxy_for_aws/utils.py
@@ -18,10 +18,9 @@ import argparse
 import httpx
 import logging
 import os
-import re
 from fastmcp.client.transports import StreamableHttpTransport
 from mcp_proxy_for_aws.sigv4_helper import create_sigv4_client
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple
 from urllib.parse import urlparse
 
 
@@ -72,8 +71,38 @@ def create_transport_with_sigv4(
     )
 
 
+def get_service_name_and_region_from_endpoint(endpoint: str) -> Tuple[str, str]:
+    """Extract service name and region from an endpoint URL.
+
+    Args:
+        endpoint: The endpoint URL to parse
+
+    Returns:
+        Tuple of (service_name, region). Either value may be empty string if not found.
+
+    Notes:
+        - Matches bedrock-agentcore endpoints (gateway and runtime)
+        - Matches AWS API Gateway endpoints (service.region.api.aws)
+        - Falls back to extracting first hostname segment as service name
+    """
+    # Parse AWS service from endpoint URL
+    parsed = urlparse(endpoint)
+    hostname = parsed.hostname or ''
+    match hostname.split('.'):
+        case [*_, 'bedrock-agentcore', region, 'amazonaws', 'com']:
+            return 'bedrock-agentcore', region  # gateway and runtime
+        case [service, region, 'api', 'aws']:
+            return service, region
+        case [service, *_]:
+            # Fallback: extract first segment as service name
+            return service, ''
+        case _:
+            logger.warning('Could not parse endpoint, no hostname found')
+            return '', ''
+
+
 def determine_service_name(endpoint: str, service: Optional[str] = None) -> str:
-    """Validate and determine the service name.
+    """Validate and determine the service name and possibly region from an endpoint.
 
     Args:
         endpoint: The endpoint URL
@@ -81,6 +110,7 @@ def determine_service_name(endpoint: str, service: Optional[str] = None) -> str:
 
     Returns:
         Validated service name
+        Validated region
 
     Raises:
         ValueError: If service cannot be determined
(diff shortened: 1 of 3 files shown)


tests/unit/test_utils.py
```diff
@@ -136,6 +136,19 @@ def test_validate_service_name_service_parsing_simple_hostname(self):
         result = determine_service_name(endpoint)
         assert result == 'myservice'
 
+    def test_validate_service_name_bedrock_agentcore(self):
+        """Test parsing service name for bedrock-agentcore endpoints."""
+        # Test various bedrock-agentcore endpoint formats
+        test_cases = [
```
harv-aws, line 142: this is a real agentcore runtime url - 
https://bedrock-agentcore.us-east-1.amazonaws.com/runtimes/arn%3Aaws%3Abedrock-agentcore%3Aus-east-1%3A216123401914%3Aruntime%2Fhosted_agent_99wdf-hYKYrgAGVr/invocations

Can you capture a testcase for something like above ?

harv-aws (commented):
Aligned on functionality. added a minor comment on test.

tests/unit/test_utils.py
arangatang: Note that for AgentcoreRuntime our normal `--service` detection works as intended as the beginning of the url is `bedrock-agentcore`.

mcp_proxy_for_aws/utils.py
```diff
@@ -89,8 +89,14 @@ def determine_service_name(endpoint: str, service: Optional[str] = None) -> str:
...
-    # Extract service name (first part before first dot or dash)
-    service_match = re.match(r'^([^.]+)', hostname)
+    if '.bedrock-agentcore.' in hostname:
+        # Bedrock agentcore does not follow the expected service name
+        logger.info('Bedrock agent core detected, resolving service name to bedrock-agentcore')
+        return 'bedrock-agentcore'
+    else:
+        # Extract service name (first part before first dot or dash)
+        logger.info('Resolving service name')
+        service_match = re.match(r'^([^.]+)', hostname)
```
wzxxing: I think these logic can be simplified and made more readable without using regex.

The regex is just to get the first label in a the DNS name.


```suggestion
    match hostname.split("."):
        case [gateway_id, "bedrock-agentcore", region, "amazonaws", "com"]:
            return "bedrock-agentcore" # gateway
        case ["bedrock-agentcore", region, "amazonaws", "com"]:
            return "bedrock-agentcore" # runtime 
        case [mcp_server_name, region, "api", "aws"]:
            return mcp_server_name
```

mcp_proxy_for_aws/utils.py
```diff
@@ -89,8 +89,20 @@ def determine_service_name(endpoint: str, service: Optional[str] = None) -> str:
...
 
-    # Extract service name (first part before first dot or dash)
-    service_match = re.match(r'^([^.]+)', hostname)
+    if '.bedrock-agentcore.' in hostname:
+        # Bedrock agentcore does not follow the expected service name
+        logger.info('Bedrock agent core detected, resolving service name to bedrock-agentcore')
+        return 'bedrock-agentcore'
+    else:
+        # Extract service name (first part before first dot or dash)
+        logger.info('Resolving service name')
```
wzxxing: can be removed.

wzxxing (changes requested):

mcp_proxy_for_aws/utils.py
```diff
@@ -92,9 +92,13 @@ def determine_service_name(endpoint: str, service: Optional[str] = None) -> str:
...
 
-    # Extract service name (first part before first dot or dash)
-    service_match = re.match(r'^([^.]+)', hostname)
-    determined_service = service_match.group(1) if service_match else None
+    logger.info('Resolving service name')
+    determined_service = ''
+    match hostname.split('.'):
+        case [_, 'gateway', 'bedrock-agentcore', _, 'amazonaws', 'com']:
+            determined_service = 'bedrock-agentcore'  # gateway
+        case result:
```
wzxxing: I think this `case` statement is probably wrong?

What does it do?

wzxxing: Oh, I think I understand. But maybe better to be explicit.

```python
        case [service_name, _region, "amazonaws", "com]:
            determined_service = ...
        case _:
           pass
```

wzxxing: also seems like there is a potention for us to determine the region from the endpoint too? Could you do a follow up on that?

arangatang: `[service_name, _region, "amazonaws", "com]`

This only matches that specific pattern, but when we in reality only care about the first part no matter of the suffix. Thinking someone could host their own mcp not under `amazonaws` suffix as well if they would want.

wzxxing (approved):

PR #86, opened 2025-11-13
fix(sigv4_helper.py): reduce severity of log levels

kyoncal:
## Summary
Log levels were changed.

### Changes

> Log levels were changed.

### User experience

> The user will now not see error logs when metadata injection is skipped.

## Checklist

* [] Changes have been tested
* [] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/sigv4_helper.py
+++ b/mcp_proxy_for_aws/sigv4_helper.py
@@ -250,9 +250,9 @@ async def _inject_metadata_hook(metadata: Dict[str, Any], request: httpx.Request
         metadata: Dictionary of metadata to inject into _meta field
         request: The HTTP request object
     """
-    logger.info('=== Outgoing Request ===')
-    logger.info('URL: %s', request.url)
-    logger.info('Method: %s', request.method)
+    logger.debug('=== Outgoing Request ===')
+    logger.debug('URL: %s', request.url)
+    logger.debug('Method: %s', request.method)
 
     # Try to inject metadata if it's a JSON-RPC/MCP request
     if request.content and metadata:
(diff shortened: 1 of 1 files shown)


wzxxing (approved):

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -298,4 +298,4 @@ async def _inject_metadata_hook(metadata: Dict[str, Any], request: httpx.Request
 
         except (json.JSONDecodeError, KeyError, TypeError) as e:
             # Not a JSON request or invalid format, skip metadata injection
-            logger.error('Skipping metadata injection: %s', e)
+            logger.info('Skipping metadata injection: %s', e)
```
wzxxing: Actually this should also be a `debug`
```suggestion
            logger.debug('Skipping metadata injection: %s', e)
```

wzxxing (changes requested):

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -284,7 +284,7 @@ async def _inject_metadata_hook(metadata: Dict[str, Any], request: httpx.Request
                             )
                     body['params']['_meta'] = {**metadata, **existing_meta}
                 else:
-                    logger.info('Replacing non-dict _meta value with injected metadata')
```
arangatang, line 287: Thanks for the change! Could you also add a section in the readme under troubleshooting that Cline has some issues with logging during the connection phase and count those as errors?

wzxxing: Thanks for noticing this. @kyoncal This is the weird part of the cline code for your reference.

https://github.com/cline/cline/blob/04549cf1e3b6461d78f8cd95100c7b56d8529cf4/src/services/mcp/McpHub.ts#L281-L303

PR #95, opened 2025-11-25
write initialize error to stdout

wzxxing:
### Changes

When client fails, we want to the see the failure reason in the MCP client. This was possible with v1.0 where every request is independent, but with v1.1.1, to reuse the MCP session for every request, proxy takes an already connected mcp client, so the client initialization happens outside of the MCP client life cycle.

### User experience

The user should see a more detailed error message such as below, immediately after the connection is closed:

```
Failed to connect to MCP server "aws-local": MCP error -32001: Your AWS session token has expired. Please refresh your credentials and try again. If you're using temporary credentials, you may need to re-authenticate to obtain new credentials.
```

Previously v1.1.1 make the error message very unclear:

```
Failed to connect to MCP server "AWSMCPFromLatestPyPi": MCP error -32000: Connection closed
```

### Testing
update MCP config as 
```json
"aws": {
      "command": "uvx",
      "args": [
        "--from",
        "git+https://github.com/aws/mcp-proxy-for-aws@wzxxing/write-last-error",
        "mcp-proxy-for-aws",
        "--service",
        "aws-mcp",
        "--profile",
        "mcp-tester",
        "--region",
        "us-east-1",
        "--log-level",
        "DEBUG",
        "<remote-mcp-server-url>"
      ]
    },
```

Expired credentials (Authn failure):

Screenshot 2025-11-26 at 10 24 34

Credentials without permission (Authz faillure)

image

Tool call failure (Authz), normal without terminating the transport

image

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [ ] No

## Acknowledgment

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ The MCP Proxy serves as a lightweight, client-side bridge between MCP clients (A
 uvx mcp-proxy-for-aws@latest <SigV4 MCP endpoint URL>
 ```
 
+**Note:** The first run may take tens of seconds as `uvx` downloads and caches dependencies. Subsequent runs will start in seconds. Actual startup time depends on your network and hardware.
+
+
 #### Using a local repository
 
 ```bash


mcp_proxy_for_aws/server.py
```diff
@@ -84,16 +85,29 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
...
+            add_tool_filtering_middleware(proxy, args.read_only)
+
+            if args.retries:
+                add_retry_middleware(proxy, args.retries)
+            await proxy.run_async(transport='stdio')
+    except Exception as e:
+        logger.error('Failed to start server: %s', e)
+        cause = e.__cause__
+        if isinstance(cause, McpProxy) and cause.payload:
+            print(cause.payload, file=sys.stdout)
```
wzxxing: This is to print the initialization error to the process stdout so that the client can receive it and show the error message properly.

mcp_proxy_for_aws/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class McpProxyError(Exception):
+    """Error which will be handled by the proxy writing the payload to stdout."""
+
+    def __init__(self, payload: str | None = None, *args: object) -> None:
+        """Consturct a MCP proxy error with payload."""
```
PCManticore: Typo here.

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -195,7 +196,12 @@ async def _handle_error_response(response: httpx.Response) -> None:
             # Try to parse JSON error details
             error_details = response.json()
             logger.log(log_level, 'HTTP %d Error Details: %s', response.status_code, error_details)
-        except Exception:
+            if response.status_code == 401 or response.status_code == 403:
+                request = await response.request.aread()
+                request_body = json.loads(request.decode())
+                if isinstance(request_body, dict) and request_body.get('method') == 'initialize':
```
PCManticore: Why is this thrown only on `initialize`? I presume because other MCP operations would normally return 200 if they are implemented according to the spec. But we can't make that assumption especially as the proxy can be leveraged by any non-AWS service specific MCP server (for example an AgentCore Runtime MCP server would be feasible to use with the proxy).

mcp_proxy_for_aws/server.py
```diff
@@ -84,16 +85,29 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
...
-        await proxy.run_async()
+    try:
+        async with Client(transport=transport) as client:
+            # Create proxy with the transport
+            proxy = FastMCP.as_proxy(
+                client,
+                name='MCP Proxy',
+                instructions=(
+                    'MCP Proxy for AWS Server that provides access to backend servers through a single interface. '
+                    'This proxy handles authentication and request routing to the appropriate backend services.'
```
arangatang, line 167: nit: SigV4 authentication and we don't really do any routing.

mcp_proxy_for_aws/server.py
```diff
@@ -84,16 +85,29 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
...
-
-        await proxy.run_async()
+    try:
+        async with Client(transport=transport) as client:
+            # Create proxy with the transport
+            proxy = FastMCP.as_proxy(
+                client,
+                name='MCP Proxy',
+                instructions=(
+                    'MCP Proxy for AWS Server that provides access to backend servers through a single interface. '
```
arangatang:
```suggestion
                    'MCP Proxy for AWS provides access to SigV4 protected MCP servers through a single interface. '
```

mcp_proxy_for_aws/server.py
arangatang: nit/qq: Maybe we want to do this also via the `logger` but point it to `sys.stdout` only in this case.

mcp_proxy_for_aws/server.py
```diff
@@ -84,16 +85,29 @@ async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
...
+
+            if args.retries:
+                add_retry_middleware(proxy, args.retries)
+            await proxy.run_async(transport='stdio')
+    except Exception as e:
+        logger.error('Failed to start server: %s', e)
+        cause = e.__cause__
+        if isinstance(cause, McpProxyError) and cause.payload:
+            print(cause.payload, file=sys.stdout)
+        raise e
```
arangatang: nit: raise
```suggestion
        raise
```

mcp_proxy_for_aws/sigv4_helper.py
arangatang: > would normally return 200 if they are implemented according to the spec. 

Note, this is wrong. If done by [spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management), e.g. when a SessionId is expired the server must return 404 if its passed. 
However I do agree this is not really respected properly 😓

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -195,7 +196,12 @@ async def _handle_error_response(response: httpx.Response) -> None:
             # Try to parse JSON error details
             error_details = response.json()
             logger.log(log_level, 'HTTP %d Error Details: %s', response.status_code, error_details)
-        except Exception:
+            if response.status_code == 401 or response.status_code == 403:
```
arangatang: Should we do this on any 4xx / 5xx errors per default and ignore failures for  maybe GET requests?

e.g. allowlist for know "ok" failures like GET requests instad of this surgical way?
wdyt?

wzxxing: Move the error handling outside of sigv4 helper

mcp_proxy_for_aws/server.py
```diff
@@ -37,13 +48,75 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-async def setup_mcp_mode(local_mcp: FastMCP, args) -> None:
+@contextlib.asynccontextmanager
+async def _initialize_client(transport: ClientTransport):
+    """Handle the exceptions for during client initialize."""
+    # line = sys.stdin.readline()
+    # logger.debug('First line from kiro %s', line)
```
arangatang, line 60:
```suggestion
```

arangatang (dismissed):

kyoncal (approved):

PR #97, opened 2025-11-26
fix: use factory to refresh session once it is finished

arnewouters:
### Changes

Timeouts stop the background session task in FastMCP, this makes follow-up tool calls fail with

```
MCP Tool Error Response: Internal error: nesting counter should be 0 when starting new session, got 2
``` 

With this PR, we are explicitly refreshing the session when the client is no longer connected.

### User experience

config for testing:
```json
    "example-mcp-server": {
      "command": "uv",
        "args": [
          "run",
          "--directory",
          "/Volumes/workplace/mcp-proxy-for-aws-dev",
          "mcp-proxy-for-aws",
          "--service",
          "my-service",
          "--region",
          "us-east-1",
          "--log-level",
          "DEBUG",
          "https://example.com/mcp"
        ],
        "env": {},
        "disabled": false,
        "autoApprove": []
    }
```

#### Cline

Before:

There is an error after a timeout

Untitled Diagram drawio

After:

Session refreshes under the hood and the next call succeeds

Untitled Diagram2 drawio

## Checklist

* [x] Changes are documented

* [ ] Yes
* [ ] No

## Acknowledgment

### DIFF
--- a/tests/unit/test_initialize_client.py
+++ b/tests/unit/test_initialize_client.py
@@ -28,7 +28,7 @@ async def test_successful_initialization():
     mock_transport = Mock()
     mock_client = Mock()
 
-    with patch('mcp_proxy_for_aws.server.Client') as mock_client_class:
+    with patch('mcp_proxy_for_aws.server.ProxyClient') as mock_client_class:
         mock_client_class.return_value.__aenter__ = AsyncMock(return_value=mock_client)
         mock_client_class.return_value.__aexit__ = AsyncMock(return_value=None)
 
@@ -48,7 +48,7 @@ async def test_http_error_with_jsonrpc_error(capsys):
 
     http_error = httpx.HTTPStatusError('error', request=Mock(), response=mock_response)
 
-    with patch('mcp_proxy_for_aws.server.Client') as mock_client_class:
+    with patch('mcp_proxy_for_aws.server.ProxyClient') as mock_client_class:
         mock_client_class.return_value.__aenter__ = AsyncMock(side_effect=http_error)
 
         with pytest.raises(httpx.HTTPStatusError):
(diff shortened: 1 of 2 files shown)


mcp_proxy_for_aws/server.py
```diff
@@ -157,10 +158,20 @@ async def run_proxy(args) -> None:
     transport = create_transport_with_sigv4(
         args.endpoint, service, region, metadata, timeout, profile
     )
+
     async with _initialize_client(transport) as client:
```
wzxxing, line 161: The client should be a ProxyClient.

arnewouters: Why? We are doing the same as `FastMCP.as_proxy` https://github.com/jlowin/fastmcp/blob/ba69fba3055db6938ba368dc17275a85ca626ae3/src/fastmcp/server/server.py#L2612-L2634

wzxxing: Because we probably want a mix of the if-else branches

1. connect the session to be reused
2. use ProxyClient to support features like elicitation and sampling; the regular client does not support

mcp_proxy_for_aws/server.py
```diff
@@ -157,10 +157,20 @@ async def run_proxy(args) -> None:
...
+
+        async def client_factory():
+            nonlocal client
+            if not client.is_connected():
+                logger.debug('Reinitialize client')
+                client = ProxyClient(transport)
```
wzxxing, line 167: what if you do `client.new()` here?

mcp_proxy_for_aws/server.py
```diff
@@ -157,10 +157,20 @@ async def run_proxy(args) -> None:
...
+                await client._connect()
```
wzxxing, line 168: When this method throws, the client also hangs, the `_initialize_client` context manager writes the jsonrpc error to stdout.

Can this be done here too? I think it is a bit tricky to get the new json rpc message id.

mcp_proxy_for_aws/server.py
arnewouters: It returns Client and not ProxyClient

Issue #117, opened 2025-12-08
MCP Server won't start in Kiro CLI (using IAM Identity Center AWS CLI profile)


psantus:
### Describe the bug (Security related? please follow <https://github.com/aws/mcp-proxy-for-aws/security/policy> to report them to AWS Security directly.)

I'm having trouble to set up  AWS MCP server with Kiro CLI.
```
   "aws-mcp": {
      "command": "uvx",
      "timeout": 100000,
      "transport": "stdio",
      "args": [
        "mcp-proxy-for-aws@latest",
        "https://aws-mcp.us-east-1.api.aws/mcp"
      ]
    }
```

I keep hitting
> Mcp error: -30001: The provided SessionId was not found or has expired, please re-initialize your connection.

despite 
* being logged in with aws sso login
* having AWS_PROFILE declared (sts get-caller-identity works)
* being account admin 

adding `"--metadata", "AWS_PROFILE=<myprofile>"` to the configuration doesn't solve.

### Expected Behavior

MCP starts

### Current Behavior

MCP server doesn't

### Reproduction Steps

Configure MCP server
Start Kiro CLI

### Possible Solution

-

### Additional Information/Context

Running on Kiro CLI v1.21.0 on MacOS

### Operating System

MacOS

### Release Version

aws-mcp : latest
aws-cli: latest aws-cli/2.32.3 Python/3.13.9 Darwin/24.6.0 exe/arm64

### Search verification

- [x] I have searched existing issues and verified this is not a duplicate

### Related issues

-

wzxxing:
@psantus Could you share the session ID by enabling the debug mode?

```
"aws-mcp": {
      "command": "uvx",
      "timeout": 100000,
      "transport": "stdio",
      "args": [
        "mcp-proxy-for-aws@latest",
        "https://aws-mcp.us-east-1.api.aws/mcp",
        "--profile",
        "xxxxx",
        "--log-level",
        "DEBUG"
      ]
    }
```

**Please only share the mcp-session-id header**.

wzxxing:
Thanks for confirming, will close this issue.

PR #122, opened 2025-12-09
fix: reuse the boto3 session to sign request

wzxxing:
instead of creating a new session to sign every request

### Changes

Previously, the httpx client factory creates a new aws session object to sign the request with sigv4. This works with plain credentials, but not with assume-role credentials provider, which uses a different session role name with timestamp when the credentials is assumed.

Since the owner of the MCP session is the actors User ID (in the form of `<principal-id>:session-name>`, the users creates a session (mcp `initialize`), but will fail the following up requests when trying to access the MCP session ID with a different role session name.

fixes: https://github.com/aws/mcp-proxy-for-aws/issues/117

### User experience

User should be able to user their aws profile with assume role credentials provider.

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

Tested with a profile with assume-role credentials

## Acknowledgment


arnewouters (approved):

psantus:
I'll test when this is released, thanks for addressing my bug report!

PR #126, opened 2025-12-11
feat: build and publish container image

detti456:
### Changes

Container images are built and published on release to https://gallery.ecr.aws/mcp-proxy-for-aws/mcp-proxy-for-aws, as part of the GitHub workflow.

Software Bill of Materials (SBOM) are generated as part of the release workflow and saved as an artifact. They can also be generated through a script manually.

### User experience

Users can use the pre-built image directly from the public AWS ECR registry.

## Checklist

* [X] Changes are documented

* [ ] Yes
* [X] No

## Acknowledgment


README.md
```diff
@@ -151,6 +165,29 @@ Add the following configuration to your MCP client config file (e.g., for Amazon
...
+{
+  "mcpServers": {
+    "<mcp server name>": {
+      "command": "docker",
+      "args": [
+        "run",
+        "--rm",
+        "--volume",
+        "/full/path/to/.aws:/app/.aws:ro",
+        "public.ecr.aws/awslabs-mcp/aws/mcp-proxy-for-aws:latest",
```
arangatang: We should not use `awslabs-mcp` i'll need to figure out something new here.

.github/workflows/pypi-publish-on-release.yml
arangatang, line 1: Given this PR is out of date, can you ensure there is no drift between the two files. Maybe for now lets avoid renaming them to ensure nothing is lost.

examples/mcp-client/llamaindex/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "mcp-client-example-llamaindex"
 version = "0.1.0"
-requires-python = ">=3.10,<3.14"
```
arangatang: Do we have to do this here?

pyproject.toml
```diff
@@ -14,7 +14,7 @@ version = "1.1.5"
 
 description = "MCP Proxy for AWS"
 readme = "README.md"
-requires-python = ">=3.10,<3.14"
```
arangatang: Again, is this needed in this pr?

README.md
```diff
@@ -71,6 +71,20 @@ uv run mcp_proxy_for_aws/server.py <SigV4 MCP endpoint URL>
...
+Docker images are published to the [public AWS ECR registry](https://gallery.ecr.aws/awslabs-mcp).
+
+You can use the pre-built image:
+
+```bash
+# Pull the latest image
+docker pull public.ecr.aws/awslabs-mcp/aws/mcp-proxy-for-aws:latest
+
+# Or pull a specific version
+docker pull public.ecr.aws/awslabs-mcp/aws/mcp-proxy-for-aws:1.1.4
```
arangatang: Let's ensure these commands work after the public ecr is up. Also we need to update the url accordingly `awslabs` is not the target.

.github/workflows/pypi-publish-on-release.yml
```diff
@@ -1,4 +1,4 @@
-name: Publish to PyPI
+name: Publish to PyPI and ECR
```
wzxxing: Is it possible to separate these two actions? one publish to pypi, one publish to ECR.

Dockerfile
```diff
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# dependabot should continue to update this to the latest hash.
-FROM public.ecr.aws/docker/library/python:3.14.0-alpine3.22@sha256:8373231e1e906ddfb457748bfc032c4c06ada8c759b7b62d9c73ec2a3c56e710 AS uv
+# Using Amazon Linux for consistency and compliance
+FROM public.ecr.aws/amazonlinux/amazonlinux:latest AS uv
```
wzxxing: In general, I wonder why this dockerfile does not use the newly published mcp-proxy-for-aws pypi package directly by doing a pip install and that's all.

Are all the gcc compilers required?

If the resulted image size is too large, you could use mounted cache with the yum package cache directories.

https://docs.docker.com/build/cache/optimize/#use-cache-mounts

.github/workflows/pypi-publish-on-release.yml
detti456: I separated them, the ECR workflow now depends on the PyPI workflow

Dockerfile
detti456: I changed it, the Dockerfile is simpler now. Publishing to ECR can only start after the package is successfully published to PyPI but I think that's not an issue.

Dockerfile
```diff
@@ -12,79 +12,29 @@
...
 
 # Install runtime dependencies and create application user
 RUN yum update -y && \
     yum install -y \
         python3.13 \
+        python3.13-pip \
         ca-certificates \
-        shadow-utils && \
+        shadow-utils \
+        lsof && \
```
arangatang, line 30: Do we need `lsof`?

Dockerfile
```diff
@@ -12,79 +12,29 @@
...
     yum clean all && \
     update-ca-trust && \
     groupadd -r app && \
     useradd -r -g app -d /app app
 
-# Copy application artifacts from build stage
-COPY --from=uv --chown=app:app /app/.venv /app/.venv
+# Install mcp-proxy-for-aws from PyPI
+RUN python3.13 -m pip install mcp-proxy-for-aws
```
arangatang, line 37: Did you verify that the created image works in e.g. Kiro with this change?

Dockerfile
detti456: Yes, for the docker health checks to work properly.

Dockerfile
detti456: Yes, I tested the application entrypoint and it works as expected.

PR #132, opened 2025-12-29
fix: Use streamable_http_client to avoid DeprecationWarning

### LINKED ISSUE #128: DeprecationWarning: Use streamable_http_client instead of streamablehttp_client

When running the application with Python 3.13, a `DeprecationWarning` is triggered indicating that the deprecated `streamablehttp_client` function is being used instead of the newer `streamable_http_client` function.

**Warning Message:**
```
DeprecationWarning: Use streamable_http_client instead
  File "contextlib.py", line 109
```

**Root Cause:**
The MCP Python SDK recently renamed and updated its HTTP client API, introducing `streamable_http_client` (with underscore) to replace the older `streamablehttp_client` (without underscore). The codebase appears to be using the deprecated function, which now emits a deprecation warning.

https://github.com/aws/mcp-proxy-for-aws/blob/main/mcp_proxy_for_aws/client.py#L117

** Reference **
https://github.com/modelcontextprotocol/python-sdk/pull/1177

### Expected Behavior

The library should use the updated streamable_http_client API without triggering any deprecation warnings.

### Current Behavior

Print Warning Message:
```
DeprecationWarning: Use streamable_http_client instead
  File "contextlib.py", line 109
```

### Reproduction Steps

Use mcp-proxy-for-aws in python like:
```
        stdio_mcp_client = MCPClient(
            lambda: stdio_client(
                StdioServerParameters(
                	command=command="uvx",
            		args=[
                		"mcp-proxy-for-aws@1.1.4",
                		"--metadata", "AWS_REGION=us-west-2"
            		]
            	)
            ),
        )``
```

### Possible Solution

_No response_

_No response_

### Operating System

Ubuntu 22.04

### Release Version

1.1.4

### Other information

_No response_

### Search verification

### Related issues

None

ps06756:
## Summary

- Updated the import in `client.py` to use the new `streamable_http_client` function name instead of the deprecated `streamablehttp_client`
- Added backward-compatible import that first tries the new function name, then falls back to the old one for older MCP SDK versions
- Updated unit tests to mock the new function name

## Details

The MCP Python SDK v1.24.0 (released Dec 12, 2025) introduced a new `streamable_http_client` function and deprecated the old `streamablehttp_client` (see [PR #1177](https://github.com/modelcontextprotocol/python-sdk/pull/1177)).

Users running with MCP SDK >= 1.24.0 see the following warning:
```
DeprecationWarning: Use streamable_http_client instead
```

This fix uses a try/except import pattern to:
1. Use the new `streamable_http_client` when available (MCP >= 1.24.0)
2. Fall back to `streamablehttp_client` for older versions (MCP < 1.24.0)

## Test plan

- [x] All existing unit tests pass (`uv run pytest tests/unit/test_client.py -v`)
- [x] Verified backward compatibility with MCP 1.20.0
- [x] Verified the new function exists in MCP 1.25.0 on PyPI

Fixes #128

### DIFF
--- a/mcp_proxy_for_aws/client.py
+++ b/mcp_proxy_for_aws/client.py
@@ -18,7 +18,12 @@ from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStre
 from botocore.credentials import Credentials
 from contextlib import _AsyncGeneratorContextManager
 from datetime import timedelta
-from mcp.client.streamable_http import GetSessionIdCallback, streamablehttp_client
+from mcp.client.streamable_http import GetSessionIdCallback
+
+try:
+    from mcp.client.streamable_http import streamable_http_client
+except ImportError:
+    from mcp.client.streamable_http import streamablehttp_client as streamable_http_client
 from mcp.shared._httpx_utils import McpHttpClientFactory, create_mcp_http_client
 from mcp.shared.message import SessionMessage
 from mcp_proxy_for_aws.sigv4_helper import SigV4HTTPXAuth
@@ -114,7 +119,7 @@ def aws_iam_streamablehttp_client(
     auth = SigV4HTTPXAuth(creds, aws_service, region)
 
     # Return the streamable HTTP client context manager with AWS IAM authentication
-    return streamablehttp_client(
+    return streamable_http_client(
         url=endpoint,
         headers=headers,
         timeout=timeout,
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -60,7 +60,7 @@ async def test_boto3_session_parameters(
     mock_read, mock_write, mock_get_session = mock_streams
 
     with patch('boto3.Session', return_value=mock_session) as mock_boto:
-        with patch('mcp_proxy_for_aws.client.streamablehttp_client') as mock_stream_client:
+        with patch('mcp_proxy_for_aws.client.streamable_http_client') as mock_stream_client:
             mock_stream_client.return_value.__aenter__ = AsyncMock(
                 return_value=(mock_read, mock_write, mock_get_session)
             )
@@ -94,7 +94,7 @@ async def test_sigv4_auth_is_created_and_used(mock_session, mock_streams, servic
 
     with patch('boto3.Session', return_value=mock_session):
         with patch('mcp_proxy_for_aws.client.SigV4HTTPXAuth') as mock_auth_cls:
-            with patch('mcp_proxy_for_aws.client.streamablehttp_client') as mock_stream_client:
+            with patch('mcp_proxy_for_aws.client.streamable_http_client') as mock_stream_client:
                 mock_auth = Mock()
                 mock_auth_cls.return_value = mock_auth
                 mock_stream_client.return_value.__aenter__ = AsyncMock(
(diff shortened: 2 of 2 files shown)


ps06756:
@bidesh / @harv-aws Bumping this up, for review

mcp_proxy_for_aws/client.py
```diff
@@ -114,7 +119,7 @@ def aws_iam_streamablehttp_client(
     auth = SigV4HTTPXAuth(creds, aws_service, region)
 
     # Return the streamable HTTP client context manager with AWS IAM authentication
-    return streamablehttp_client(
+    return streamable_http_client(
```
bidesh, line 122: My understanding from this change https://github.com/modelcontextprotocol/python-sdk/pull/1177 is that the signature has changed on how the client is passed.

If `streamable_http_client` is imported instead of streamablehttp_client, I think this will break because the new implementation will use a default http_client which will not have the headers, auth etc here set up.

bidesh (changes requested):
@ps06756 Thanks for the change!

I added a comment regarding the compatibility of the two methods. I think the tests might pass either because you were still getting streamablehttp_client imported (because the mcp included is still old) or we have gaps in our tests. 

Tests we will look after, but if you could verify that the change works are expected without regression, then happy to approve and merge. Thanks for the change!

mcp_proxy_for_aws/client.py
ps06756: Yes, good catch! I had wrongly assumed that we would be using the version > 1.24.0 always, 

In this case, how do you suggest we determine whether we are on a new version or old version ? 

One option is we can use the `packaging` dependency and do something like 
```
from importlib.metadata import version
  from packaging.version import Version

  # Check version at import time
  _MCP_VERSION = Version(version('mcp'))
  _NEW_API_AVAILABLE = _MCP_VERSION >= Version('1.24.0')

  if _NEW_API_AVAILABLE:
      from mcp.client.streamable_http import streamable_http_client as _streamable_http_client_new
  else:
      from mcp.client.streamable_http import streamablehttp_client as _streamablehttp_client_old
```

Or we can directly check the method signature at runtime and determine that ? 

```
import inspect

  _NEW_API_AVAILABLE = False
  _streamable_http_client_new = None
  _streamablehttp_client_old = None

  try:
      from mcp.client.streamable_http import streamable_http_client as _candidate

      sig = inspect.signature(_candidate)
      # New API has 'http_client' param, old API has 'auth' param
      if 'http_client' in sig.parameters and 'auth' not in sig.parameters:
          _NEW_API_AVAILABLE = True
          _streamable_http_client_new = _candidate
  except (ImportError, ValueError):
      pass

  if not _NEW_API_AVAILABLE:
      try:
          from mcp.client.streamable_http import streamablehttp_client as _streamablehttp_client_old
      except ImportError:
          from mcp.client.streamable_http import streamable_http_client as _streamablehttp_client_old
```

bidesh: Actually, the only way we get rid of the DeprecationWarning is if we update the version. I talked to @wzxxing on any problems with upgrading. Copying his reply here: 

> The fastmcp v2.14.1 introduced some incompatible change that changes the signature of the the httpx client factory.
> The fix should be just to add kw args to this function.
https://github.com/aws/mcp-proxy-for-aws/blob/1bea695d2b0d6f0223d0b33988a70fdb6aa90378/mcp_proxy_for_aws/utils.py#L60
[3:28](https://amzn-aws.slack.com/archives/C06HJ7MJZ9S/p1768228089165859?thread_ts=1768226488.601479&cid=C06HJ7MJZ9S)

@ps06756 would you be willing to include this change and upgrade the dependency? 

If not, then please comment in the linked issue and we will prioritise the fix and update.

PR #135, opened 2026-01-07
chore: using updated aws_iam_streamable_http_client

andreyyoshuamanik:
## Summary

This pull request introduces a new, improved AWS IAM streamable HTTP client function, deprecates the old function, and updates the package to align with upstream MCP (Managed Control Plane) conventions. The changes modernize the API, improve consistency with upstream libraries, and enhance test coverage. There are also minor improvements to environment variable handling and dependency requirements.

#### API Modernization & Deprecation

* Added a new `aws_iam_streamable_http_client` function to replace the deprecated `aws_iam_streamablehttp_client`, aligning with upstream MCP naming and signature patterns. The new function uses keyword-only arguments, accepts a pre-configured `httpx.AsyncClient`, and drops several legacy parameters. The return type is now `AsyncGenerator` for consistency. (`mcp_proxy_for_aws/client.py`, `CHANGELOG.md`) [[1]](diffhunk://#diff-d8d9f6919601c49790e6c77957bbf7e9953b374f3d4cdb193a613c5b0e0a73e3R16-R43) [[2]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR8-R27)
* Marked `aws_iam_streamablehttp_client` as deprecated; it now wraps the new function for backward compatibility and will be removed in a future major release. (`mcp_proxy_for_aws/client.py`, `CHANGELOG.md`) [[1]](diffhunk://#diff-d8d9f6919601c49790e6c77957bbf7e9953b374f3d4cdb193a613c5b0e0a73e3R126-R228) [[2]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR8-R27)

#### Dependency and Version Updates

* Updated the minimum required version of `fastmcp` to 2.14.2 to support the new client interface and MCP features. (`pyproject.toml`, `CHANGELOG.md`) [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L13-R19) [[2]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR8-R27)

#### Testing and Internal Refactoring

* Refactored all tests to use the new `aws_iam_streamable_http_client` and updated mocks to patch `streamable_http_client` instead of the old function. Tests now verify the correct use of HTTP client factories, parameter passing, and backward compatibility of the deprecated API. (`tests/unit/test_client.py`) [[1]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL20-R20) [[2]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL63-R63) [[3]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL97-R104) [[4]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bR114-R129) [[5]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL140-R153) [[6]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bR166-R193) [[7]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL173-R206) [[8]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL186-R222) [[9]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL201-R237) [[10]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL223-R259) [[11]](diffhunk://#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL267-R303)

#### Utilities and Environment Handling

* Improved environment variable handling in integration tests: `AWS_SESSION_TOKEN` is now only set if present in credentials, supporting both permanent and temporary credentials. (`tests/integ/mcp/simple_mcp_client.py`)
* Updated the client factory utility to accept additional keyword arguments for compatibility with new `fastmcp` features. (`mcp_proxy_for_aws/utils.py`)

### User experience

Before: there will be a deprecation message because using deprecated streamablehttp_client
After: no deprecation message anymore

## Checklist

* [x] Changes are documented

* [x] Yes
* [ ] No

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/client.py
+++ b/mcp_proxy_for_aws/client.py
@@ -13,12 +13,14 @@
 # limitations under the License.
 
 import boto3
+import httpx
 import logging
+import warnings
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from botocore.credentials import Credentials
 from contextlib import _AsyncGeneratorContextManager
 from datetime import timedelta
-from mcp.client.streamable_http import GetSessionIdCallback, streamablehttp_client
+from mcp.client.streamable_http import GetSessionIdCallback, streamable_http_client
 from mcp.shared._httpx_utils import McpHttpClientFactory, create_mcp_http_client
 from mcp.shared.message import SessionMessage
 from mcp_proxy_for_aws.sigv4_helper import SigV4HTTPXAuth
@@ -28,7 +30,7 @@ from typing import Optional
 logger = logging.getLogger(__name__)
 
 
-def aws_iam_streamablehttp_client(
+def aws_iam_streamable_http_client(
     endpoint: str,
     aws_service: str,
     aws_region: Optional[str] = None,
@@ -60,7 +62,7 @@ def aws_iam_streamablehttp_client(
         credentials: Optional AWS credentials from boto3/botocore. If provided, takes precedence over aws_profile.
         headers: Optional additional HTTP headers to include in requests.
         timeout: Request timeout in seconds or timedelta object. Defaults to 30 seconds.
-        sse_read_timeout: Server-sent events read timeout in seconds or timedelta object.
+        sse_read_timeout: Deprecated. This parameter is no longer used and will be removed in version 2.0.0.
         terminate_on_close: Whether to terminate the connection on close.
         httpx_client_factory: Factory function for creating HTTPX clients.
 
@@ -71,7 +73,7 @@ def aws_iam_streamablehttp_client(
             - get_session_id: Callback function to retrieve the current session ID
 
     Example:
-        async with aws_iam_mcp_client(
+        async with aws_iam_streamable_http_client(
             endpoint="https://example.com/mcp",
             aws_service="bedrock-agentcore",
             aws_region="us-west-2"
@@ -80,6 +82,13 @@ def aws_iam_streamablehttp_client(
             pass
     """
     logger.debug('Preparing AWS IAM MCP client for endpoint: %s', endpoint)
+
+    # Warn if sse_read_timeout is set to a non-default value
+    if sse_read_timeout != 60 * 5:
+        logger.warning(
+            'sse_read_timeout parameter is deprecated and will be removed in version 2.0.0. '
+            'The value is ignored in the current implementation.'
+        )
 
     if credentials is not None:
         creds = credentials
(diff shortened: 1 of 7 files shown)


mcp_proxy_for_aws/client.py
```diff
@@ -28,7 +28,7 @@
 logger = logging.getLogger(__name__)
 
 
-def aws_iam_streamablehttp_client(
+def aws_iam_streamable_http_client(
```
arnewouters: This will be a breaking change for existing users

andreyyoshuamanik: Actually that is intended, to align with the streamable_http_client syntax. And i guess it will only break users that update the package version (?) please cmiiw

andreyyoshuamanik:
> The direction is right - upstream deprecated `streamablehttp_client`. But this change needs a deprecation cycle.
> 
> Keep `aws_iam_streamablehttp_client` as a deprecated alias that wraps the new function and emits a warning. Then bump to minor version, add a CHANGELOG entry, and plan removal for the next major.
> 
> Upstream [did this same pattern](https://github.com/modelcontextprotocol/python-sdk/blob/3863f203e91cbb2b21e7cfc24ed8e6d3520b62c7/src/mcp/client/streamable_http.py#L685). Let's not break existing code without a migration window.

Makes sense! Thanks for the suggestion. Adjusted the approach

mcp_proxy_for_aws/client.py
```diff
@@ -113,13 +122,74 @@ def aws_iam_streamablehttp_client(
...
+    """Create an AWS IAM-authenticated MCP streamable HTTP client.
+
+    .. deprecated:: 1.2.0
+        Use :func:`aws_iam_streamable_http_client` instead.
+        This function will be removed in version 2.0.0.
+
+    This is a deprecated alias for aws_iam_streamable_http_client.
+    Please update your code to use aws_iam_streamable_http_client instead.
+    """
+    warnings.warn(
```
DennisTraub: I suggest using the `@deprecated` decorator:

```
from typing_extensions import deprecated

...

@deprecated("Use `aws_iam_streamable_http_client` instead.")
def aws_iam_streamablehttp_client(
    ...
```

mcp_proxy_for_aws/client.py
```diff
@@ -28,7 +30,7 @@
 logger = logging.getLogger(__name__)
 
 
-def aws_iam_streamablehttp_client(
+def aws_iam_streamable_http_client(
```
DennisTraub: If we change the API to match the upstream naming convention, we should also match the signature. It wouldn't make much sense to introduce a breaking change just to keep naming consistent while letting the signature drift.

```
async def aws_iam_streamable_http_client(
    # Specific to the IAM client, replacing upstream's `url: str`
    endpoint: str,
    aws_service: str,
    aws_region: str | None = None,
    aws_profile: str | None = None,
    credentials: Credentials | None = None,
    # Remaining upstream parameters following the new pattern
    *,
    http_client: httpx.AsyncClient | None = None,
    terminate_on_close: bool = True,
) -> AsyncGenerator[
    tuple[
        MemoryObjectReceiveStream[SessionMessage | Exception],
        MemoryObjectSendStream[SessionMessage],
        GetSessionIdCallback,
    ],
    None,
]:
    ...
```

mcp_proxy_for_aws/client.py
andreyyoshuamanik: Agree. updated in this commit ya https://github.com/aws/mcp-proxy-for-aws/pull/135/commits/24f4bccff8b24f5f1fff8377f997a48e6640861f

mcp_proxy_for_aws/client.py
andreyyoshuamanik: This makes sense. Thanks for point this out. i match the signature with the upstream in this commit https://github.com/aws/mcp-proxy-for-aws/pull/135/commits/24f4bccff8b24f5f1fff8377f997a48e6640861f

PR #205, opened 2026-03-24
feat(auth): add per-call AWS profile override middleware

### LINKED ISSUE #119: Impossible to use 2 distinct profiles in a single Kiro CLI session.

I set up AWS_PROFILE=profile1

then I launch Kiro CLI, set up with 
```
"mcpServers": {
    "aws-mcp": {
      "command": "uvx",
      "transport": "stdio", 
      "timeout": 100000,
      "args": [
        "mcp-proxy-for-aws@latest",
        "https://aws-mcp.us-east-1.api.aws/mcp"
      ]
    }
}
```

I ask "get caller identity". 

KiroCLI runs
```
Running tool aws___call_aws with the param (from mcp server: aws-mcp)
 ⋮  {
 ⋮    "cli_command": "aws sts get-caller-identity"
 ⋮  }
```

then I say "do the same with AWS_PROFILE=profile2

KrioCLI runs
```
Running tool aws___call_aws with the param (from mcp server: aws-mcp)
 ⋮  {
 ⋮    "cli_command": "aws sts get-caller-identity --profile profile2"
 ⋮  }
```
but the result returned is exactly the same (though profiles are using different roles in different accounts): the MCP sever persists in using the session that was set up with Kiro CLI started. 

### Expected Behavior

LLM should be able to use multiple CLI profiles.

### Current Behavior

The profile used when Kiri CLI launches is the only one that get all the queries.

### Reproduction Steps

See above.

### Possible Solution

_No response_

_No response_

### Operating System

MacOS

### Release Version

latest (this server, Kiro-cli and AWS CLI all in latest version)

### Other information

_No response_

### Search verification

### Related issues

-

### LINKED ISSUE #172: switch_profile tool to allow multi-aws account access
### Describe the feature (Security related? please follow <https://github.com/aws/mcp-proxy-for-aws/security/policy> to report them to AWS Security directly.)

Hello!

The feature request is to give this MCP functionality to allow agents to assume profiles for multi-account access.

### Use Case

I was using steampipe in order to accomplish this before, but that is only helpful for some read-only operations. When I used this MCP, I needed to close Claude Code, assume a new profile, relaunch Claude Code, and repeat X amount of times depending on what I'm doing. We have 70 AWS accounts so you can imagine why this might be a pain.

### Proposed Solution

I have a very scrappy tool that Ive created which enabled me to do this already, if its something you think could be a good addition to this project I can tidy it up and create a PR. Let me know what you think!

The switch_profile tool lets you change which AWS identity the proxy uses mid-session. You enable it with --allow-switch-profile and a whitelist of profile names. When called, a middleware intercepts the request, validates the profile against the allowlist, creates a new boto3 session, swaps it into a shared SessionHolder, reconnects the upstream client, and invalidates the tool cache. All subsequent requests are then signed with the new credentials automatically.

### Other Information

_No response_

### Acknowledgements

- [x] I may be able to implement this feature request
- [x] This feature might incur a breaking change

### LINKED ISSUE #193: Support credential refresh when switching AWS profiles/roles without proxy restart
### Describe the feature (Security related? please follow <https://github.com/aws/mcp-proxy-for-aws/security/policy> to report them to AWS Security directly.)

Currently, the proxy creates a single `boto3.Session` at startup in `create_transport_with_sigv4()` (utils.py) and reuses it for the lifetime of the process. While `session.get_credentials()` is called on every request in `_sign_request_hook()`, botocore caches the resolved credentials in the Session object after the first read from `~/.aws/credentials`.

This means that if a user switches to a different AWS account/role locally (e.g., by changing `AWS_PROFILE`, updating `~/.aws/credentials`, running `aws sso login` with a different profile, or modifying environment variables), the proxy continues to use the stale credentials from the original session. The only workaround is to fully restart the proxy process.

### Use Case

When working with multiple AWS accounts (e.g., dev, staging, production), developers frequently switch between different IAM roles or SSO profiles during a single session. Common approaches include:

- Switching named profiles with `role_arn` in `~/.aws/config` (via `AWS_PROFILE`)
- Using AWS SSO / IAM Identity Center with different profiles (`aws sso login --profile <profile>`)
- Using `aws login` and switching profiles
- Updating environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`)

In all of these cases, the proxy does not pick up the new credentials because the `boto3.Session` was created at startup and its credential provider cache is never invalidated.

This forces users to restart the MCP proxy (and in some MCP clients, restart the entire chat session) every time they switch accounts. This is disruptive to workflow, especially during development sessions where cross-account access is needed frequently.

### Proposed Solution

[Might be other better ways] Auto-detect the local change of creds, and then tear down the old connection between the proxy and the remote mcp server, and then re-establish a connection. 

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

benjstoll:
Adds ProfileOverrideMiddleware that allows routing individual tool calls through dedicated per-profile MCP connections via a `profile` argument. Enabled with `--allow-switch-profile` CLI flag restricted to an explicit allowlist of profile names.

### Changes

  - Added ProfileOverrideMiddleware in mcp_proxy_for_aws/middleware/profile_switcher.py that intercepts a profile argument on any tool call,
  validates it against an allowlist, and routes the request through a dedicated per-profile MCP client with its own SigV4-signed transport.
  - Added --allow-switch-profile CLI argument in cli.py that accepts one or more AWS profile names to enable the middleware.
  - Wired the middleware into server.py with proper lifecycle management (lazy client creation, graceful shutdown of per-profile connections
  in the finally block).
  - Added 12 unit tests covering tool schema injection, pass-through behavior, disallowed profiles, argument stripping, connection/tool-call
  error handling, and client disconnect logic. 90% branch coverage on the new middleware.
  - Updated README.md with the new parameter in the configuration table and a "Multi-account access" section explaining how
  --allow-switch-profile interacts with --profile, including a JSON config example.

### User experience

  Before: Users who needed to query AWS resources across multiple accounts had to run separate proxy instances per profile, or manually
  restart the proxy with a different --profile value.

  After: Users pass --allow-switch-profile profile-a profile-b alongside their default --profile. Any tool call can include a profile
  argument to route that single request through a dedicated connection signed with the specified profile's credentials. Tool calls without
  profile continue to use the default connection. Each profile's connection is created lazily on first use, so there is no startup cost for
  unused profiles.

## Checklist

* [x] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment


mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,191 @@
...
+        arguments = context.message.arguments
+        if isinstance(arguments, dict) and 'profile' in arguments:
+            profile = arguments['profile']
+            return await self._call_with_profile(profile, context, call_next)
+
+        return await call_next(context)
+
+    # ── internals ─────────────────────────────────────────────────
+
+    async def _get_profile_client(self, profile: str) -> Client:
```
anasstahr, line 129: Fastmcp supports concurrent tool calling: https://github.com/PrefectHQ/fastmcp/blob/f01d0c581c7a821a9701d6dde4d9beb95e32d479/docs/updates.mdx#L69

> ⚡ Concurrent Sampling — context.sample() can now execute multiple tool calls in parallel with tool_concurrency=0.

Should we add a lock when getting the profile client?
```
  def __init__(self, ...):
      ...
      self._profile_clients: dict[str, Client] = {}
      self._lock = asyncio.Lock()

  async def _get_profile_client(self, profile: str) -> Client:
      async with self._lock:
          if profile not in self._profile_clients:
              logger.info('Creating dedicated connection for profile %s', profile)
              transport = create_transport_with_sigv4(...)
              client = Client(transport=transport)
              await client.__aenter__()
              self._profile_clients[profile] = client
          return self._profile_clients[profile]
```

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,191 @@
...
+        """Get or create a dedicated MCP client for the given profile.
+
+        Each profile gets its own ``StreamableHttpTransport`` and MCP session
+        so that requests signed with different AWS identities don't collide
+        on the same backend session.
+        """
+        if profile not in self._profile_clients:
+            logger.info('Creating dedicated connection for profile %s', profile)
+            transport = create_transport_with_sigv4(
```
anasstahr: Should we also have a test for calling `get_profile_client`, checking this code path?

mcp_proxy_for_aws/middleware/profile_switcher.py
benjstoll: that's a good idea, this would prevent a race condition with two async calls to the same profile. I pushed your suggestion with an added test case.

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,194 @@
...
+    @override
+    async def on_call_tool(
+        self,
+        context: MiddlewareContext[mt.CallToolRequestParams],
+        call_next: CallNext[mt.CallToolRequestParams, ToolResult],
+    ) -> ToolResult:
+        """Intercept ``profile`` and route through a dedicated per-profile client."""
+        arguments = context.message.arguments
+        if isinstance(arguments, dict) and 'profile' in arguments:
+            profile = arguments['profile']
```
rshevchuk-git: `profile` can collide with tools that already have a profile parameter. If a backend tool already has a profile parameter in its schema, this middleware will silently overwrite it in on_list_tools and strip it in on_call_tool. This could break legitimate tool parameters.

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,194 @@
...
+    async def _call_with_profile(
+        self,
+        profile: str,
+        context: MiddlewareContext[mt.CallToolRequestParams],
+        call_next: CallNext[mt.CallToolRequestParams, ToolResult],
+    ) -> ToolResult:
+        """Forward a tool call through a dedicated per-profile connection."""
+        if profile not in self._allowed_profiles:
+            allowed = ', '.join(sorted(self._allowed_profiles))
+            return ToolResult(
```
rshevchuk-git: Why do you return ToolResult(s) on failed validations instead of raising ToolError?

mcp_proxy_for_aws/server.py
```diff
@@ -100,13 +102,27 @@ async def run_proxy(args) -> None:
...
+            profile_middleware = ProfileOverrideMiddleware(
+                allowed_profiles=allowed_profiles,
+                service=service,
+                region=region,
+                metadata=metadata,
+                timeout=timeout,
+                endpoint=args.endpoint,
+            )
+            proxy.add_middleware(profile_middleware)
+
```
rshevchuk-git: This should be extracted into a separate helper `add_profile_override_middleware`, to stay consistent with other middleware declarations

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,194 @@
...
+        """Inject ``profile`` into every tool's schema."""
+        tools = await call_next(context)
+
+        for tool in tools:
+            params = tool.parameters
+            if not isinstance(params, dict):
+                continue
+            if 'properties' not in params:
+                params['properties'] = {}
+            params['properties']['profile'] = {
```
rshevchuk-git: If the upstream tool list is cached or shared, this mutates shared state. This is ok now because AWSProxyToolManager.get_tools() re-fetches, however this creates dependency on the upstream to always produce fresh dicts. I'd suggest deep-copy the parameters dict before mutating. It is cheap insurance and makes the middleware self-contained.

rshevchuk-git (changes requested):

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,194 @@
...
+    @override
+    async def on_list_tools(
+        self,
+        context: MiddlewareContext[mt.ListToolsRequest],
+        call_next: CallNext[mt.ListToolsRequest, Sequence[Tool]],
+    ) -> Sequence[Tool]:
+        """Inject ``profile`` into every tool's schema."""
+        tools = await call_next(context)
+
+        for tool in tools:
```
arnewouters, line 87: Curious if we could use [tool transformations](https://gofastmcp.com/servers/transforms/tool-transformation) instead of directly manipulating params?

mcp_proxy_for_aws/middleware/profile_switcher.py
benjstoll: You're right, I should have caught this, ill return ToolError here.

mcp_proxy_for_aws/server.py
benjstoll: Thanks, ill make sure this matches the other helpers.

mcp_proxy_for_aws/middleware/profile_switcher.py
benjstoll: Do you think proxy_profile would be a better name to avoid collisions?

mcp_proxy_for_aws/middleware/profile_switcher.py
benjstoll: agreed

mcp_proxy_for_aws/middleware/profile_switcher.py
benjstoll: Im not sure this is a good fit because our proxy_profile parameter doesnt exist in any backend tool, we are just injecting it so the middleware can route calls to specific AWS profiles. Let me know what your thoughts were exactly for tool transformations.

mcp_proxy_for_aws/middleware/profile_switcher.py
anasstahr: I think `proxy_profile` sounds good, however I think we should warn the user when this override happens

mcp_proxy_for_aws/middleware/profile_switcher.py
arnewouters: I haven't used this feature before but it seems like it's only for removing parameters and not adding them as we are doing here. You can ignore this :)

mcp_proxy_for_aws/middleware/profile_switcher.py
benjstoll: Sounds good to me, added a logger.warning if a collision is detected.

anasstahr:
Hi @benjstoll, could you fix the build errors? We can go ahead with merging this also please run the integration tests and confirm that they're passing

benjstoll:
> Hi @benjstoll, could you fix the build errors? We can go ahead with merging this also please run the integration tests and confirm that they're passing

Looks like it was a bad import, I went ahead and fixed that. Integration tests seems to be passing from my end!

rorymcmahon:
This would unblock us from adopting the managed AWS MCP server. We have 40+ accounts in an Identity Center-managed landing zone and the single-profile-per-session limitation is currently the only reason we're staying on the local `awslabs.aws-api-mcp-server`.

The lazy connection creation per-profile and the allowlist approach are both great design choices — no startup cost for unused profiles, and the allowlist gives us a security boundary to prevent agents from accidentally touching production accounts without explicit permission.

Looking forward to this landing.

benjstoll:
Hi @anasstahr, anything else? Looks like this might be good to go

aswhit1:
+1, but we have 118 AWS accounts we're working across!

rshevchuk-git:
We have added multi profile support in #266, release is going to happen soon.

PR #233, opened 2026-04-08
fix: remove fast fail credential check

anasstahr:
Removes the early credential validation check in `create_aws_session` that called `session.get_credentials()` and raised a `ValueError` if no credentials were immediately available. Updates unit tests to reflect the new behavior.

### Changes

  - Removed the early credential validation check in `create_aws_session` that called `session.get_credentials()` and raised a `ValueError` if no credentials were immediately available
  - Updated unit tests to reflect the new behavior

### User Experience

  Before: The proxy would fail immediately at startup if AWS credentials weren't resolvable at session creation time.

  After: The session is created without checking credentials — failures surface at request-signing time.

## Checklist

* [x] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment


mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -100,22 +100,13 @@ def create_aws_session(profile: Optional[str] = None) -> boto3.Session:
...
         session = boto3.Session(profile_name=profile) if profile else boto3.Session()
     except Exception as e:
         raise ValueError(f"Failed to create AWS session with profile '{profile}': {e}")
 
-    # Verify credentials are available
-    credentials = session.get_credentials()
-    if not credentials:
-        profile_msg = f" with profile '{profile}'" if profile else ''
-        raise ValueError(
-            f'No AWS credentials found{profile_msg}. '
```
krishmakochhar, line 116: Is this method credential check only happening at session creation, or is there also validation at signing time? If it's only here, removing it means users with missing credentials won't get a helpful error at all — just botocore's raw exception. Is there similar handling downstream?

anasstahr: There is validation downstream at signing time — `_sign_request_hook` calls `session.get_credentials()` on every request. 
For authenticated endpoint we return 403, which is caught by `ToolErrorMiddleware` - 401/403 errors, so we won't return a raw exception

arnewouters (approved):

rshevchuk-git (approved):

PR #245, opened 2026-04-14
fix: refresh stale credentials on auth failure without restart

anasstahr:
When AWS credentials expire, the proxy now automatically picks up refreshed credentials on the next request without requiring a restart.

Problem:
- boto3.Session was created once at startup and cached forever
- session.get_credentials() returned the same stale Credentials object
- Even after refreshing creds on disk (e.g. aws sso login), the proxy kept using the old frozen session until restarted

Fix:
- Add SessionHolder that wraps the boto3 session with lazy refresh
- On 401/403, mark the session for refresh (don't refresh immediately, since creds on disk may not be updated yet)
- On the next request's signing, create a fresh boto3.Session that reads the current credentials from disk
- Improve error messages: credential errors now clearly say 'expired or invalid AWS credentials' instead of 'Unknown tool'

The lazy refresh pattern ensures the new session is created at signing time (after the user has refreshed creds), not at error time (when creds may still be stale).

### Manual credential expiry test

Used the following script to verify end-to-end: get credentials, connect to the AWS MCP server, wait 20 minutes for expiry, refresh credentials, and make another request.

```python
"""Manual test for credential refresh across session expiry."""

import asyncio
import fastmcp
import logging
import subprocess
import sys
from fastmcp.client import StdioTransport

logging.basicConfig(level=logging.INFO, format='%(asctime)s %(levelname)s %(message)s')

for _name in ('fastmcp', 'mcp', 'httpx', 'httpcore', 'botocore', 'urllib3'):
    logging.getLogger(_name).setLevel(logging.CRITICAL)

REFRESH_COMMAND = ['your-credential-refresh-command', '--once']
ENDPOINT = 'https://aws-mcp.us-east-1.api.aws/mcp'
REGION = 'us-east-1'
WAIT_MINUTES = 20


def refresh_credentials():
    logger.info('Running: %s', ' '.join(REFRESH_COMMAND))
    result = subprocess.run(REFRESH_COMMAND, capture_output=True, text=True)
    if result.returncode != 0:
        logger.error('Credential refresh failed (exit %d):\n%s', result.returncode, result.stderr)
        sys.exit(1)
    logger.info('Credentials refreshed')


async def make_request(client: fastmcp.Client):
    await client.ping()
    logger.info('Ping OK')
    tools = await client.list_tools()
    logger.info('Listed %d tools', len(tools))
    response = await client.call_tool('aws___list_regions', {})
    assert response.is_error is False, f'Tool call failed: {response}'
    logger.info('list_regions call succeeded')


async def main():
    refresh_credentials()
    client = fastmcp.Client(
        StdioTransport(
            command='mcp-proxy-for-aws',
            args=[ENDPOINT, '--region', REGION, '--log-level', 'CRITICAL'],
        ),
        timeout=30.0,
    )
    async with client:
        logger.info('--- First request (fresh credentials) ---')
        await make_request(client)
        logger.info('Waiting %d minutes for credentials to expire...', WAIT_MINUTES)
        await asyncio.sleep(WAIT_MINUTES * 60)
        refresh_credentials()
        logger.info('--- Second request (after credential refresh) ---')
        await make_request(client)
    logger.info('Test passed: both requests succeeded')


if __name__ == '__main__':
    asyncio.run(main())
```

Result:

```
2026-04-14 14:16:44,956 INFO list_regions call succeeded
2026-04-14 14:16:44,957 INFO Test passed: both requests succeeded
```

### Changes

### User experience

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [ ] No

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/middleware/tool_error_middleware.py
+++ b/mcp_proxy_for_aws/middleware/tool_error_middleware.py
@@ -62,17 +62,32 @@ class ToolErrorMiddleware(Middleware):
             logger.error('Tool call %r failed: %s.', tool_name, e)
             message = f'Tool call {tool_name!r} failed: {e}. Please retry.'
             if self._is_credential_error(e):
-                message += (
-                    ' This may be caused by expired or invalid AWS credentials.'
-                    ' Consider using long-lived credentials such as an AWS profile'
-                    ' (--profile) or IAM Identity Center (aws sso login).'
+                message = (
+                    f'Tool call {tool_name!r} failed due to expired or invalid AWS credentials.'
+                    ' Please refresh your credentials and retry.'
+                    ' The proxy will automatically use the new credentials on the next request.'
                 )
             raise ToolError(message) from e
 
     @staticmethod
     def _is_credential_error(error: Exception) -> bool:
         """Check if the error is likely caused by expired or invalid credentials."""
-        return isinstance(error, httpx.HTTPStatusError) and error.response.status_code in (
-            401,
-            403,
-        )
+        # Walk the exception chain — the 401/403 may be wrapped
+        current: BaseException | None = error
+        while current is not None:
+            if isinstance(current, httpx.HTTPStatusError) and current.response.status_code in (
+                401,
+                403,
+            ):
+                return True
+            current = current.__cause__ if current.__cause__ else current.__context__
+            # Avoid infinite loops on self-referencing chains
+            if current is error:
+                break
+
+        # "Unknown tool" after a failed reconnect is almost always a credential issue
+        error_str = str(error)
+        if 'Unknown tool' in error_str or 'Unauthorized' in error_str:
+            return True
+
+        return False
--- a/mcp_proxy_for_aws/middleware/tool_error_middleware.py
+++ b/mcp_proxy_for_aws/middleware/tool_error_middleware.py
@@ -71,8 +71,12 @@ class ToolErrorMiddleware(Middleware):
 
     @staticmethod
     def _is_credential_error(error: Exception) -> bool:
-        """Check if the error is likely caused by expired or invalid credentials."""
-        # Walk the exception chain — the 401/403 may be wrapped
+        """Check if the error is likely caused by expired or invalid credentials.
+
+        Walks the exception chain (__cause__/__context__) because the
+        HTTPStatusError may be wrapped.  isinstance already respects the
+        MRO, so subclasses are caught too.
+        """
         current: BaseException | None = error
         while current is not None:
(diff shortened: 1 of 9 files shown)


mcp_proxy_for_aws/middleware/tool_error_middleware.py
```diff
@@ -62,17 +62,24 @@ async def on_call_tool(
...
     @staticmethod
     def _is_credential_error(error: Exception) -> bool:
         """Check if the error is likely caused by expired or invalid credentials."""
-        return isinstance(error, httpx.HTTPStatusError) and error.response.status_code in (
-            401,
-            403,
-        )
+        # Walk the exception chain — the 401/403 may be wrapped
+        current: BaseException | None = error
+        while current is not None:
```
arnewouters: I think we can make this cleaner with MRO (e.g. `error.mro()`)

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -258,14 +289,24 @@ async def _sign_request_hook(
...
+    session_holder.refresh_if_needed()
+
     # Get AWS credentials from the session
-    credentials = session.get_credentials()
+    credentials = session_holder.session.get_credentials()
+    logger.debug(
+        'Signing request with credentials for access key: %s...%s',
+        credentials.access_key[:4],
+        credentials.access_key[-4:],
+    )
```
arnewouters: We just removed this in #242

arnewouters (dismissed):
Can you check the coverate of the tests manually? `uv run pytest --cov --cov-report html`, let's make sure all new lines are covered :)

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -120,12 +120,42 @@ def create_aws_session(profile: Optional[str] = None) -> boto3.Session:
...
+
+    def refresh_if_needed(self) -> None:
+        """Create a fresh session if a previous request got an auth error."""
+        if not self._needs_refresh:
+            return
+        logger.info('Refreshing AWS session to pick up new credentials')
+        try:
+            self.session = create_aws_session(self._profile)
+            self._needs_refresh = False
+        except ValueError:
```
rshevchuk-git: Why do we only catch ValueError?

mcp_proxy_for_aws/middleware/tool_error_middleware.py
anasstahr: I don't think so, we would still need to walk the exception chain, pretty sure mro won't help with this

arnewouters (approved):

rshevchuk-git (approved):

PR #260, opened 2026-04-28
feat: add --skip-auth flag to opt-in to unsigned requests

anasstahr:
### Changes

Adds a `--skip-auth` CLI flag (off by default) that allows the proxy to send unsigned requests when AWS credentials are unavailable. Without this flag, the proxy now raises a clear error when credentials cannot be resolved, guiding users to either configure credentials or explicitly opt in to unsigned requests.

### User experience

**Before:** The proxy would silently send unsigned requests whenever AWS credentials were unavailable.

**After:** The proxy raises a `ValueError` with a helpful message when credentials are unavailable. Users who intentionally want to skip signing can pass `--skip-auth` to allow unsigned requests to unauthenticated endpoints.

```bash
# Default: requires credentials
mcp-proxy-for-aws https://my-endpoint.com/mcp

# Opt-in to unsigned requests
mcp-proxy-for-aws https://my-endpoint.com/mcp --skip-auth
```

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

- Unit tests updated for `_sign_request_hook` covering: raising when credentials are None (default), skipping signing with `skip_auth=True`, and not instantiating `SigV4HTTPXAuth`
- Integration tests pass (16/16) against local MCP server

## Acknowledgment


mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -294,6 +294,10 @@ async def _sign_request_hook(
     # Get AWS credentials from the session
     credentials = session_holder.session.get_credentials()
 
+    if credentials is None:
+        logger.debug('No AWS credentials available, sending request unsigned')
+        return
```
arnewouters: Will we end up here if there is a non-existing profile configured?

anasstahr: Code is altered, but we will get:
```
ValueError: Failed to create AWS session with profile 'bogus': The config profile (bogus) could not be found
```

arnewouters (approved):

rshevchuk-git (approved):

PR #266, opened 2026-05-01
feat(auth): add per-call AWS profile override middleware

anasstahr:
Based on the work from #205 by @benjstoll. Adds ProfileOverrideMiddleware that allows routing individual tool calls through dedicated per-profile MCP connections via an `aws_profile` tool parameter.

### Changes

- Added `ProfileOverrideMiddleware` in `mcp_proxy_for_aws/middleware/profile_switcher.py` that intercepts an `aws_profile` argument on any tool call, validates it against an allowlist, and routes the request through a dedicated per-profile MCP client with its own SigV4-signed transport.
- `--profile` accepts multiple values. First is the default, rest enable per-call switching (e.g., `--profile default dev staging`).
- Wired the middleware into `server.py` with proper lifecycle management (lazy client creation, graceful shutdown of per-profile connections in the finally block).
- Added unit tests covering tool schema injection, pass-through behavior, disallowed profiles, argument stripping, connection/tool-call error handling, and client disconnect logic.
- Updated README.md with the new parameter in the configuration table and a "Multi-account access" section.

### Additional changes on top of #205

- Unified `--profile` to accept multiple values, removed `--allow-switch-profile` flag
- Renamed `proxy_profile` to `aws_profile` — the agent doesn't need to know about the proxy
- Added `ToolError` raises on failures instead of silent fallbacks
- Added asyncio lock to prevent duplicate profile client creation
- Added warning when a tool's `aws_profile` parameter is shadowed by the override
- Added dedup logic: duplicate profiles removed, default excluded from switch list
- Fixed pyright compatibility for fastmcp.tools import path
- Used `dest='profiles'` in argparse for clarity

## Closes

- #172
- #119
- #193

### DIFF
--- a/mcp_proxy_for_aws/middleware/profile_switcher.py
+++ b/mcp_proxy_for_aws/middleware/profile_switcher.py
@@ -29,8 +29,8 @@ import logging
 import mcp.types as mt
 from collections.abc import Sequence
 from fastmcp import Client
-from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
 from fastmcp.exceptions import ToolError
+from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
 from fastmcp.tools.tool import Tool, ToolResult
 from mcp_proxy_for_aws.utils import create_transport_with_sigv4
 from typing import Any, cast
--- a/mcp_proxy_for_aws/middleware/profile_switcher.py
+++ b/mcp_proxy_for_aws/middleware/profile_switcher.py
@@ -31,7 +31,7 @@ from collections.abc import Sequence
 from fastmcp import Client
 from fastmcp.exceptions import ToolError
 from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
-from fastmcp.tools.tool import Tool, ToolResult
+from fastmcp.tools import Tool, ToolResult
 from mcp_proxy_for_aws.utils import create_transport_with_sigv4
 from typing import Any, cast
 from typing_extensions import override
--- a/mcp_proxy_for_aws/middleware/profile_switcher.py
+++ b/mcp_proxy_for_aws/middleware/profile_switcher.py
@@ -1,0 +1,191 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Middleware that enables per-call AWS profile overrides via a ``profile`` argument.
+
+Pass ``profile`` as an extra argument on any tool call to route that single request
+through a dedicated transport signed with the specified profile's credentials. The
+argument is stripped before forwarding to the backend.
+
+Each profile gets its own lazily-created ``StreamableHttpTransport`` and MCP session,
+so parallel subagents querying different accounts don't interfere with each other.
+"""
+
+import httpx
+import logging
+import mcp.types as mt
+from collections.abc import Sequence
+from fastmcp import Client
+from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
+from fastmcp.tools.tool import Tool, ToolResult
+from mcp_proxy_for_aws.utils import create_transport_with_sigv4
+from typing import Any, cast
+from typing_extensions import override
+
+
+logger = logging.getLogger(__name__)
+
+
+class ProfileOverrideMiddleware(Middleware):
+    """Middleware that intercepts ``profile`` on any tool call for per-request AWS identity switching.
+
+    When a tool call includes a ``profile`` argument, the middleware:
+
+    1. Validates the profile against the allowed list
+    2. Strips ``profile`` from the arguments
+    3. Forwards the call through a dedicated per-profile MCP client
+
+    Each profile gets its own transport and session to the backend so that
+    requests signed with different AWS identities don't collide.
+    """
+
+    def __init__(
+        self,
+        allowed_profiles: list[str],
+        service: str,
+        region: str,
+        metadata: dict[str, Any],
+        timeout: httpx.Timeout,
+        endpoint: str,
+    ) -> None:
+        """Initialize the middleware with connection and profile configuration."""
+        super().__init__()
+        self._allowed_profiles = set(allowed_profiles)
+        self._endpoint = endpoint
+        self._service = service
+        self._region = region
+        self._metadata = metadata
+        self._timeout = timeout
+        self._profile_clients: dict[str, Client] = {}
+
+    # ── tool listing ────────────────────────────────────────────────
+
+    @override
+    async def on_list_tools(
+        self,
+        context: MiddlewareContext[mt.ListToolsRequest],
+        call_next: CallNext[mt.ListToolsRequest, Sequence[Tool]],
+    ) -> Sequence[Tool]:
+        """Inject ``profile`` into every tool's schema."""
+        tools = await call_next(context)
+
+        for tool in tools:
+            params = tool.parameters
+            if not isinstance(params, dict):
+                continue
+            if 'properties' not in params:
+                params['properties'] = {}
+            params['properties']['profile'] = {
+                'type': 'string',
+                'description': (
+                    'AWS CLI profile to sign this request with. Omit to use the default profile.'
+                ),
+                'enum': sorted(self._allowed_profiles),
+            }
+
+        return list(tools)
+
+    # ── tool invocation ─────────────────────────────────────────────
+
+    @override
+    async def on_call_tool(
+        self,
+        context: MiddlewareContext[mt.CallToolRequestParams],
+        call_next: CallNext[mt.CallToolRequestParams, ToolResult],
+    ) -> ToolResult:
+        """Intercept ``profile`` and route through a dedicated per-profile client."""
+        arguments = context.message.arguments
+        if isinstance(arguments, dict) and 'profile' in arguments:
+            profile = arguments['profile']
+            return await self._call_with_profile(profile, context, call_next)
+
+        return await call_next(context)
+
+    # ── internals ─────────────────────────────────────────────────
+
+    async def _get_profile_client(self, profile: str) -> Client:
+        """Get or create a dedicated MCP client for the given profile.
+
+        Each profile gets its own ``StreamableHttpTransport`` and MCP session
+        so that requests signed with different AWS identities don't collide
+        on the same backend session.
+        """
+        if profile not in self._profile_clients:
+            logger.info('Creating dedicated connection for profile %s', profile)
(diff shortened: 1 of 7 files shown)


mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,203 @@
...
+            params = copy.deepcopy(tool.parameters)
+            if 'properties' not in params:
+                params['properties'] = {}
+            if 'proxy_aws_profile' in params['properties']:
+                logger.warning(
+                    'Tool %r already defines a "proxy_aws_profile" parameter; '
+                    'the middleware override is shadowing the backend definition.',
+                    tool.name,
+                )
+            params['properties']['proxy_aws_profile'] = {
```
rshevchuk-git: This injects proxy_aws_profile into every tool's schema, which contradicts the proposal and token-reduction tenet. What is the motivation?

mcp_proxy_for_aws/server.py
```diff
@@ -63,16 +65,23 @@ async def run_proxy(args) -> None:
     if args.metadata:
         metadata.update(args.metadata)
 
-    # Get profile
-    profile = args.profile
+    # Get profile(s) from CLI args or env var
+    profiles: list[str] = args.profiles if args.profiles else []
+    if not profiles:
```
rshevchuk-git: cli.py:99 defaults args.profiles to [AWS_PROFILE] if that env var is set. Then here you do

```
  profiles: list[str] = args.profiles if args.profiles else []
  if not profiles:
      env_profiles = os.environ.get('AWS_MCP_PROXY_PROFILES')
      ...
```

A plugin user who already has AWS_PROFILE=foo exported in their shell will never see AWS_MCP_PROXY_PROFILES honored, because args.profiles is already populated with ['foo'].

mcp_proxy_for_aws/server.py
```diff
@@ -63,16 +65,23 @@ async def run_proxy(args) -> None:
...
+        env_profiles = os.environ.get('AWS_MCP_PROXY_PROFILES')
+        if env_profiles:
+            profiles = env_profiles.split()
+    default_profile = profiles[0] if profiles else None
+    # Dedup switch profiles and exclude the default profile
+    switch_profiles = list(dict.fromkeys(p for p in profiles[1:] if p != default_profile))
```
rshevchuk-git: If config is `--profile default dev staging`, the schema enum is [dev, staging]. If an agent decides to pass `proxy_aws_profile=default` because user commanded to use 'default' profile - it will get a ToolError. The schema description hints "Omit to use the default profile", but real-world agents might copy enum values verbatim. I'd suggest include the default in the enum and treat proxy_aws_profile=<default> as the no-override path.

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,203 @@
...
+    async def disconnect_profile_clients(self) -> None:
+        """Disconnect all per-profile clients. Call during server shutdown."""
+        for profile, client in self._profile_clients.items():
+            try:
+                await client.__aexit__(None, None, None)
+            except Exception:
+                logger.exception('Failed to disconnect profile client %s', profile)
+        self._profile_clients.clear()
+
+    async def _call_with_profile(
```
rshevchuk-git, line 190: ProfileOverrideMiddleware._call_with_profile calls client.call_tool(...) directly and never invokes call_next(context). That means retries are skipped for profile-overridden calls. Default-profile calls retry on transient errors; overridden calls don't.

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,203 @@
...
+        except Exception:
+            logger.exception('Failed to create connection for profile %s', profile)
+            raise ToolError(
+                f'Failed to create connection for profile {profile!r}. '
+                'Check that the profile is configured and credentials are valid.'
+            )
+
+        try:
+            result = await client.call_tool(context.message.name, arguments)
+            return ToolResult(
```
rshevchuk-git, line 236: I think we're loosing isError here.

mcp_proxy_for_aws/server.py
```diff
@@ -63,16 +65,23 @@ async def run_proxy(args) -> None:
...
 
     # Log server configuration
     logger.info(
```
rshevchuk-git, line 85: Should we also log the switch list for debugging?

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,203 @@
...
+
+    # ── tool listing ────────────────────────────────────────────────
+
+    @override
+    async def on_list_tools(
+        self,
+        context: MiddlewareContext[mt.ListToolsRequest],
+        call_next: CallNext[mt.ListToolsRequest, Sequence[Tool]],
+    ) -> Sequence[Tool]:
+        """Inject ``proxy_aws_profile`` into every tool's schema."""
```
arnewouters: Why are we using `proxy_aws_profile` over `aws_profile`?

rshevchuk-git (changes requested):

mcp_proxy_for_aws/middleware/profile_switcher.py
krishmakochhar: Updated,  For non-default profiles, `_call_with_profile` raises `ToolError(...) from e` preserving the original `ConnectionError`/`TimeoutError` in `__cause__`. `RetryMiddleware._should_retry()` explicitly inspects `__cause__` and retries on transient errors. Added tests confirming this behavior.

mcp_proxy_for_aws/middleware/profile_switcher.py
krishmakochhar: fixed

mcp_proxy_for_aws/server.py
krishmakochhar: Fixed. `AWS_MCP_PROXY_PROFILES` now takes precedence over args.profiles (which may come from `AWS_PROFILE`). The logic in server.py checks the env var first. Added `TestFix2EnvVarPrecedence` to cover this.

mcp_proxy_for_aws/server.py
krishmakochhar: fixed

mcp_proxy_for_aws/middleware/profile_switcher.py
krishmakochhar: discussed offline, brought back `aws_profile`

mcp_proxy_for_aws/middleware/profile_switcher.py
krishmakochhar: Fixed. Now only injected into `AUTH_REQUIRING_TOOLS`

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,234 @@
...
+        """Get or create a dedicated MCP client for the given profile.
+
+        Each profile gets its own ``StreamableHttpTransport`` and MCP session
+        so that requests signed with different AWS identities don't collide
+        on the same backend session.
+        """
+        async with self._lock:
+            if profile not in self._profile_clients:
+                logger.info('Creating dedicated connection for profile %s', profile)
+                transport = create_transport_with_sigv4(
```
rshevchuk-git, line 166: for this session we're losing args.disable_telemetry and args.skip_auth (see other transport creation)

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,234 @@
...
+
+    @override
+    async def on_call_tool(
+        self,
+        context: MiddlewareContext[mt.CallToolRequestParams],
+        call_next: CallNext[mt.CallToolRequestParams, ToolResult],
+    ) -> ToolResult:
+        """Intercept ``aws_profile`` and route through a dedicated per-profile client."""
+        arguments = context.message.arguments
+        if isinstance(arguments, dict) and 'aws_profile' in arguments:
```
rshevchuk-git, line 141: We never check if it's a AUTH_REQUIRING_TOOL. So if agent hallucinates profile argument on a non-auth tool we would still route it through a profile client

mcp_proxy_for_aws/server.py
```diff
@@ -63,16 +65,27 @@ async def run_proxy(args) -> None:
     if args.metadata:
         metadata.update(args.metadata)
 
-    # Get profile
-    profile = args.profile
+    # Get profile(s) from CLI args or env var
+    # AWS_MCP_PROXY_PROFILES takes precedence over --profile/AWS_PROFILE
+    env_profiles = os.environ.get('AWS_MCP_PROXY_PROFILES')
```
rshevchuk-git, line 73: We should document this variable

rshevchuk-git (changes requested):

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,234 @@
...
+        self._endpoint = endpoint
+        self._service = service
+        self._region = region
+        self._metadata = metadata
+        self._timeout = timeout
+        self._profile_clients: dict[str, Client] = {}
+        self._lock = asyncio.Lock()
+
+    # Tools that require AWS authentication and support profile switching
+    AUTH_REQUIRING_TOOLS = {
```
arnewouters, line 85: Missing `aws___get_tasks`

mcp_proxy_for_aws/server.py
```diff
@@ -63,16 +65,27 @@ async def run_proxy(args) -> None:
...
     # Log server configuration
     logger.info(
-        'Using service: %s, region: %s, metadata: %s, profile: %s',
+        'Using service: %s, region: %s, metadata: %s, profile: %s, switch profiles: %s',
         service,
         region,
         metadata,
-        profile,
+        default_profile,
+        all_profiles[1:] if len(all_profiles) > 1 else [],
```
arnewouters: We can remove the if else here, `all_profiles[1:]` returns an empty list when len is <= 0

```py
>>> a
[]
>>> a[1:]
[]
```

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -0,0 +1,234 @@
...
+    3. Forwards the call through a dedicated per-profile MCP client
+
+    Each profile gets its own transport and session to the backend so that
+    requests signed with different AWS identities don't collide.
+    """
+
+    def __init__(
+        self,
+        allowed_profiles: list[str],
+        default_profile: str,
```
arnewouters, line 61: Do we really need both?

krishmakochhar: Yes — allowed_profiles is for validation/enum, default_profile identifies which one routes through call_next (reusing the existing transport) vs creating a dedicated client. Without it we'd duplicate the default connection.

mcp_proxy_for_aws/middleware/profile_switcher.py
krishmakochhar: fixed

mcp_proxy_for_aws/middleware/profile_switcher.py
krishmakochhar: added

mcp_proxy_for_aws/middleware/profile_switcher.py
krishmakochhar: added

rshevchuk-git (approved):

PR #268, opened 2026-05-04
feat(server): add streamable-http transport option

chiting:
### Changes

Allow proxy to use `streamable-http` transport in addition to `stdio`.

### User experience

My use case is I can't use stdio because my agent is containerized and I don't want the agent accessing my aws credentials.

User can pass the `--transport` flag to use `streamable-http` in addition to `stdio` and call the MCP via HTTP.

Because of the lack of authentication, which can be implemented separately, this should be used with caution. 

## Checklist

* [x] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment


tkm112345:
This feature would be really useful for my use case too — a containerized agent that can't use stdio and shouldn't have direct access to AWS credentials. 
It's a shame this was auto-closed by the stale bot without any review.
Would maintainers consider reviving it, or is there a preferred approach for HTTP-based transport? 
Thanks @chiting for putting this together.

Issue #159, opened 2026-01-30
All tools are filtered when using --read-only with AWS MCP Server


SatoshiMoriyama:
### Describe the bug (Security related? please follow <https://github.com/aws/mcp-proxy-for-aws/security/policy> to report them to AWS Security directly.)

When using the `--read-only` flag with AWS MCP Server (`https://aws-mcp.us-east-1.api.aws/mcp`), all tools are filtered out and no tools are available.

I believe this is happening because the AWS MCP Server does not set `readOnlyHint=true` annotation on any of its tools. 

### Expected Behavior

When using the `--read-only` flag, read-only tools should be available. According to the [official documentation](https://docs.aws.amazon.com/aws-mcp/latest/userguide/understanding-mcp-server-tools.html), the following tools perform read-only operations:

- `aws___search_documentation` - Search AWS documentation
- `aws___read_documentation` - Read AWS documentation pages
- `aws___recommend` - Get content recommendations
- `aws___list_regions` - List AWS regions
- `aws___get_regional_availability` - Check regional availability
- `aws___suggest_aws_commands` - Get AWS API descriptions

### Current Behavior

All tools are filtered out with the following log messages:

```
2026-01-31 06:48:08 | INFO | mcp_proxy_for_aws.middleware.tool_filter | Filtering tools for read only: True
2026-01-31 06:48:08 | INFO | mcp_proxy_for_aws.middleware.tool_filter | Skipping tool aws___call_aws needing write permissions
2026-01-31 06:48:08 | INFO | mcp_proxy_for_aws.middleware.tool_filter | Skipping tool aws___get_regional_availability needing write permissions
2026-01-31 06:48:08 | INFO | mcp_proxy_for_aws.middleware.tool_filter | Skipping tool aws___list_regions needing write permissions
2026-01-31 06:48:08 | INFO | mcp_proxy_for_aws.middleware.tool_filter | Skipping tool aws___read_documentation needing write permissions
2026-01-31 06:48:08 | INFO | mcp_proxy_for_aws.middleware.tool_filter | Skipping tool aws___recommend needing write permissions
2026-01-31 06:48:08 | INFO | mcp_proxy_for_aws.middleware.tool_filter | Skipping tool aws___retrieve_agent_sop needing write permissions
2026-01-31 06:48:08 | INFO | mcp_proxy_for_aws.middleware.tool_filter | Skipping tool aws___search_documentation needing write permissions
[...]

PR #275, opened 2026-05-09
fix: harden read-only heuristic with naming convention comment and deny-list placeholder

### LINKED ISSUE #45: mcp-proxy-for-aws: --read-only disables call_aws, blocking read-only AWS APIs
### Describe the bug

Setting --read-only on mcp-proxy-for-aws reduces the proxy's advertised tool list to a single tool: aws___suggest_aws_commands. All ten other tools (aws___call_aws, aws___run_script, aws___list_regions, aws___read_documentation, aws___search_documentation, aws___get_regional_availability, aws___get_presigned_url, aws___get_tasks, aws___recommend, aws___retrieve_skill) are filtered out at
startup. Since call_aws is the only tool that runs AWS CLI commands, the proxy can't make any AWS API call in read-only mode, including reads like Get*, List*, Describe*. Several of the filtered tools (list_regions, read_documentation, search_documentation) appear to be inherently read-only or don't call AWS at all.

Note: --read-only is not documented in the toolkit README; the only public reference is the proxy's --help output.

### Expected Behavior

 Read APIs (Get*, List*, Describe*, Search*, etc.) should remain callable through call_aws when --read-only is set. The flag's help text — "Disable tools which may require write permissions" — implied to me that read-only AWS calls would still go through.

--read-only Disable tools which may require write permissions

A reasonable interpretation is "block write APIs, allow reads." It's also unclear why tools that don't perform any AWS write (e.g., list_regions, read_documentation, search_documentation) are filtered out alongside call_aws.

### Current Behavior

Two MCP server entries pointed at the same endpoint, differing only by --read-only:
  
  - Without --read-only: 11 tools advertised, including aws___call_aws and aws___run_script.
  - With --read-only: 1 tool advertised (aws___suggest_aws_commands).
  
aws___suggest_aws_commands generates text suggestions for AWS CLI commands but doesn't execute them. With no other tools available, the read-only proxy can't perform any AWS API call. The natural use case of letting an agent inspect prod safely while gating writes behind explicit authorization is unreachable with this primitive.

### Reproduction Steps

1. Configure two MCP server entries differing only by --read-only:
```
"aws-prod": {
    "command": "uvx",
    "args": [
       "mcp-proxy-for-aws@latest",
       "--metadata", "AWS_REGION=us-east-1",
       "--read-only"
     ],
     "env": { "AWS_PROFILE": "<prod-profile>" }
   }
```
  2. Restart the agent.
  3. Compare advertised tools per entry: the non-read-only entry exposes 11 tools; the --read-only entry exposes only aws___suggest_aws_commands
  4. Try invoking a read API like aws ecs list-clusters through the read-only entry. No callable tool exists to do so.

### Plugin Version

mcp-proxy-for-aws 1.4.2

### AI Assistant

Claude Code

### AI Assistant Version

2.1.132

### OS

Ubuntu 22.04.3 on WSL2 (Linux 6.6.87.2-microsoft-standard-WSL2)

### Other information

_No response_

AnnasMazhar:
Fixes aws/agent-toolkit-for-aws#45

The upstream AWS MCP Server doesn't set `readOnlyHint=true` on any tools, so `--read-only` mode filters out everything and becomes unusable. This adds name-based heuristic detection — tools starting with `list_`, `get_`, `describe_`, `read_`, `search_`, `retrieve_`, or `recommend_` pass through since these are consistently read-only in AWS APIs.

Write tools like `call_aws` are still blocked. Added a deny-list placeholder for future edge cases and a comment documenting the assumption. Long-term fix is upstream adding proper annotations.

### DIFF
--- a/mcp_proxy_for_aws/middleware/tool_filter.py
+++ b/mcp_proxy_for_aws/middleware/tool_filter.py
@@ -20,12 +20,18 @@ from fastmcp.tools import Tool
 from typing import Sequence
 
 
-# Tool name prefixes that are inherently read-only.
+# Assumes AWS API naming conventions (get_/list_/describe_ = read-only).
+# File upstream issue on awslabs/mcp for proper readOnlyHint annotations.
+
 # Remote MCP servers may not set readOnlyHint=true for their tools,
 # but tools with these naming patterns never mutate state.
 _READ_ONLY_TOOL_PREFIXES = (
     re.compile(r'^(list|read|search|get|describe|retrieve|recommend)_'),
 )
+
+# Reserved for future edge cases where a tool name matches the heuristic
+# but is actually mutating — override here to force it as write-only.
+_HEURISTIC_DENY_LIST: set[str] = set()
 
 
 class ToolFilteringMiddleware(Middleware):
--- a/mcp_proxy_for_aws/middleware/tool_filter.py
+++ b/mcp_proxy_for_aws/middleware/tool_filter.py
@@ -13,10 +13,19 @@
 # limitations under the License.
 
 import logging
+import re
 from collections.abc import Awaitable, Callable
 from fastmcp.server.middleware import Middleware, MiddlewareContext
 from fastmcp.tools import Tool
 from typing import Sequence
+
+
+# Tool name prefixes that are inherently read-only.
+# Remote MCP servers may not set readOnlyHint=true for their tools,
+# but tools with these naming patterns never mutate state.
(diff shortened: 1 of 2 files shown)


arnewouters:
Closing this as we want to solely rely on `readOnlyHint`. A fix for AWS MCP server is in flight, see my comment [here](https://github.com/aws/mcp-proxy-for-aws/issues/159#issuecomment-4403634574)

PR #294, opened 2026-05-27
fix: always read fresh credentials from disk on every request

anasstahr:
### Changes

Remove credential session caching from `SessionHolder` so that every request reads fresh credentials from disk. Previously, the proxy cached a `boto3.Session` and only refreshed it reactively after receiving a 401/403 error. Now, `SessionHolder.get_session()` always creates a new session, ensuring account switches and credential refreshes on disk take effect immediately.

- Replaced `mark_needs_refresh()` / `refresh_if_needed()` with a single `get_session()` method that always returns a fresh session
- Removed reactive 401/403 refresh logic from `_handle_error_response` (no longer needed)
- Simplified `create_transport_with_sigv4` — no longer creates an initial session at startup

### User experience

**Before:** If you switched AWS accounts or refreshed credentials on disk while the proxy was running, the proxy continued using the old cached credentials until they expired and triggered a 401/403. This could result in requests being signed with the wrong account's credentials.

**After:** Every request reads the current credentials from disk. Switching accounts or refreshing credentials takes effect on the very next request — no 401 roundtrip needed.

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/sigv4_helper.py
+++ b/mcp_proxy_for_aws/sigv4_helper.py
@@ -111,27 +111,10 @@ def create_aws_session(profile: Optional[str] = None) -> boto3.Session:
     return session
 
 
-class SessionHolder:
-    """Provides fresh boto3 sessions so every request reads current credentials from disk.
-
-    Instead of caching a session and refreshing reactively on auth errors,
-    this always creates a new session so account switches and credential
-    refreshes take effect immediately.
-    """
-
-    def __init__(self, profile: Optional[str] = None) -> None:
-        """Initialize SessionHolder with the given profile."""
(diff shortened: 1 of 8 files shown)


mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -112,34 +112,20 @@ def create_aws_session(profile: Optional[str] = None) -> boto3.Session:
 
 
 class SessionHolder:
```
arnewouters: Does this class still make sense?

arnewouters (approved):

PR #307, opened 2026-06-08
feat: make httpx connection pool size configurable via flags and env

### LINKED ISSUE #295: Hardcoded httpx pool size limits parallel tool-call fanout

`mcp_proxy_for_aws/sigv4_helper.py` constructs the `httpx.AsyncClient`
that serves all MCP tool calls with a hardcoded
`httpx.Limits(max_keepalive_connections=1, max_connections=5)`. There
is no CLI flag, environment variable, or config setting that exposes
these limits to operators.

For MCP clients that issue parallel tool-call fanout (e.g.,
[opencode](https://github.com/anomalyco/opencode) running parallel
`aws*` tool calls in a single agent turn), a mixed-service batch of
size >=5 saturates the connection pool. We observe a per-call
failure rate that is materially higher in cold-state (no warm
keepalive socket established yet) than in warm-state, with an
empirical concurrency breakpoint between sizes 4 and 5 in cold-state
sweeps. PR #270 and PR #286 address related credential-side races
that produce some of the same client-side error signatures, but the
underlying connection-pool capping is independent of those fixes
and constrains the steady-state ceiling for parallel fanout.

This issue requests exposing `max_connections` and
`max_keepalive_connections` (and ideally the rest of `httpx.Limits`)
as configurable -- environment variable, CLI flag, or both -- so
that operators with parallel-fanout MCP clients can size the pool
to their actual concurrency.

### Expected Behavior

A user running `mcp-proxy-for-aws` should be able to set the httpx
connection pool size to match their client's concurrency profile.
For example, an environment variable `MCP_PROXY_MAX_CONNECTIONS`
(or CLI flag `--max-connections`) read in `sigv4_helper.py` and
passed into the `httpx.Limits(...)` construction, with the current
values (5 / 1) preserved as defaults for backward compatibility.

Operators with single-tool-call clients see no change. Operators
with parallel-fanout clients can raise the limit (typical agent
turns we see issue 4-8 concurrent tool calls; some go higher) and
empirically eliminate the connection-pool saturation failure mode.

### Current Behavior

`sigv4_helper.py` (current main, line numbers vary by commit):

```python
httpx.Limits(max_keepalive_connections=1, max_connections=5)
```

is hardcoded. There is no public knob.

In our environment (downstream, opencode MCP client, managed
`aws-mcp.us-east-1.api.aws` endpoint, AWS SSO profile,
mcp-proxy-for-aws v1.5.0), parallel mixed-service fanout at
batch size >=5 produces intermittent per-call failures with
auth-style error messages (`InvalidClientTokenId`, `InvalidToken`,
`UnrecognizedClientException`, `AuthFailure`, `502 from runtime`).
A controlled n=64 CloudTrail audit confirmed all failed calls
share the proxy-layer signature (`AwsMcpEvent` present in
CloudTrail, `additionalEventData.downstreamRequests` absent --
the proxy received the tool call but never forwarded to AWS).

Same-service fanout (e.g., 10x `iam list-roles` in parallel) is
clean across n=27 calls in the same environment -- consistent
with HTTP/2 multiplexing letting same-service requests share a
single keepalive socket, while different SigV4 service signatures
appear to force separate connections and hit the pool ceiling.

We have evidence that #270 (asyncio.to_thread for credential
resolution) and #286 (asyncio.Lock around credential refresh)
will close the credential-race subset of failures we observe.
Even after those land, the connection-pool ceiling itself is a
separate, additive constraint -- a client that issues 8 parallel
mixed-service calls will still queue 3 of them behind the pool
even with credential races eliminated. Pool sizing is the right
knob for that case.

### Reproduction Steps

The connection-pool cap can be observed without any AWS
dependency by inspecting the AsyncClient construction. For an
end-to-end reproduction of the parallel-fanout saturation, the
shape we see is:

1. Configure `mcp-proxy-for-aws` v1.5.0 against
   `aws-mcp.us-east-1.api.aws/mcp` with an AWS SSO profile
2. Drive the proxy with an MCP client that issues parallel
   tool calls in a single turn -- 5+ mixed AWS services
   (e.g., `iam list-roles`, `sts get-caller-identity`,
   `s3api list-buckets`, `ec2 describe-vpcs`,
   `cloudtrail lookup-events`)
3. Observe intermittent per-call failures concentrated when
   batch size exceeds the pool capacity, particularly in
   cold-state (first batch after proxy start, before any
   keepalive socket is established)
4. Audit CloudTrail with `EventName=CallReadWriteTool` and
   confirm failed calls have `AwsMcpEvent` present and
   `additionalEventData.downstreamRequests` absent

Same-service fanout (10x calls to a single AWS service) does
NOT trigger the failure -- a useful negative control.

We can share the bench script we use for parallel-fanout sweeps
if helpful. CloudTrail eventID reproducers for our environment
are on PR #270 (n=64 audit) and PR #286 (cluster-shape subset).

### Possible Solution

Read pool sizing from environment variables (or CLI flags) in
`sigv4_helper.py`, with the current values as defaults:

```python
import os

max_connections = int(
    os.environ.get("MCP_PROXY_MAX_CONNECTIONS", "5")
)
max_keepalive_connections = int(
    os.environ.get("MCP_PROXY_MAX_KEEPALIVE_CONNECTIONS", "1")
)

httpx.Limits(
    max_connections=max_connections,
    max_keepalive_connections=max_keepalive_connections,
)
```

Equivalent CLI flags would also work; environment variables are
simpler for managed-runtime deployments where editing argv is
awkward. Either or both are fine.

Defaults preserved -> zero behavior change for existing users.
New knob -> parallel-fanout clients can size to their workload.

Relationship to existing work:

- **#270** (PR, open): moves blocking `get_credentials()` off the
  asyncio event loop via `asyncio.to_thread`. Closes the
  event-loop-blocked side of the failure mode. Independent of
  this issue
- **#286** (PR, open): adds `asyncio.Lock` to
  `SessionHolder.async_refresh_if_needed` to prevent concurrent
  refresh races. Closes the credential-rotation cluster-shape
  failure. Also independent of this issue
- This issue: connection-pool sizing. Even with #270 and #286
  merged, a client that issues N parallel mixed-service calls
  with N > pool_size will queue (N - pool_size) of them and may
  experience timeout-shaped failures rather than the
  credential-shaped failures #270/#286 address

We do not believe this issue is a security concern. The pool
cap is a performance/scalability constraint, not a privilege
boundary. Filing here per the bug-report template; happy to
relocate if there's a feature-request-flavored template that
fits better.

We're a downstream user (opencode MCP client, AWS managed
endpoint), not redistributing or forking the proxy. Happy to
provide more CloudTrail audit data, the bench script, or test
a candidate fix on our environment if useful.

### Operating System

Windows 11 (workstation client). Proxy runs via `uvx mcp-proxy-for-aws@latest` (Python 3.12).

### Release Version

v1.5.0 (and main as of 2026-05-25 source audit; line numbers shifted but the `httpx.Limits` cap is unchanged).

### Other information

n/a beyond what's referenced above.

### Search verification

### Related issues

#270, #286, #176

valter-silva-au:
## Problem

The single shared `httpx.AsyncClient` that serves every proxied MCP tool call is constructed with a hardcoded `httpx.Limits(max_keepalive_connections=1, max_connections=5)`, and nothing (CLI flag, env var, config) exposes it. MCP clients that fan out parallel `aws_*` tool calls per turn — Claude Code, opencode, typically 4-8 concurrent — saturate the 5-connection pool and hit intermittent failures.

## Change

Make the pool size configurable:

- New CLI flags `--max-connections` (default 5, must be `>= 1`) and `--max-keepalive-connections` (default 1, must be `>= 0`), validated via an integer range validator.
- Environment fallbacks `MCP_PROXY_MAX_CONNECTIONS` / `MCP_PROXY_MAX_KEEPALIVE_CONNECTIONS`; explicit flags take precedence over env vars.
- The two values are threaded `cli → server → create_transport_with_sigv4 → create_sigv4_client`, and through the per-profile override middleware so multi-account transports share the same pool sizing.

## Defaults / backward compatibility

Defaults are unchanged: `max_connections=5`, `max_keepalive_connections=1`. With no new flags/env, the `httpx.Limits` passed to the client is byte-for-byte identical to the previous hardcoded value. A dedicated test asserts `limits == httpx.Limits(max_keepalive_connections=1, max_connections=5)`.

## Scope

Pool-sizing knob only. This PR deliberately does **not** touch the auth-path serialization (`_sign_request_hook` / `next(auth_flow)`), which is a separate concern tracked in #270.

## Testing

- Added unit tests for the new `create_sigv4_client` params (default + override) and the CLI flags (defaults, parsing, env fallback, cli-overrides-env, range validation).
- Full suite: `uv run pytest tests/unit` → 224 → 232 passed.
- `ruff check` + `ruff format --check` clean; `pyright` reports 0 errors/warnings.
- Verified `--help` renders both flags and an end-to-end smoke test confirms env fallback and defaults resolve to `int`.

> Note: env var names (`MCP_PROXY_MAX_CONNECTIONS` / `MCP_PROXY_MAX_KEEPALIVE_CONNECTIONS`) follow the prefix used in the issue; the existing profiles env var uses `AWS_MCP_PROXY_PROFILES`. Easy to rename for consistency if you prefer.

Fixes #295

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ docker build -t mcp-proxy-for-aws .
 | `--tool-timeout`	   | Maximum seconds a tool call may take before being cancelled. When set, returns a graceful error to the agent instead of hanging indefinitely	                                                                                             | 300	                                                                    |No	|
 | `--skip-auth`         | Skip request signing when AWS credentials are unavailable instead of failing                                                                                                                                                             | `False`                                                                     |No	|
 | `--disable-telemetry` | Disables telemetry data collection                                                                                                                                                                                                      | `False`                                                                     |No	|
+| `--max-connections`   | Maximum number of concurrent HTTP connections in the shared pool used for all proxied tool calls. Increase this when your MCP client fans out parallel `aws_*` tool calls per turn (e.g. Claude Code, opencode).                          | 5 (falls back to `MCP_PROXY_MAX_CONNECTIONS`)                               |No	|
+| `--max-keepalive-connections` | Maximum number of idle keep-alive connections to retain in the pool                                                                                                                                                             | 1 (falls back to `MCP_PROXY_MAX_KEEPALIVE_CONNECTIONS`)                     |No	|
 
 ### Optional Environment Variables
 
@@ -131,9 +133,16 @@ export AWS_REGION=<aws_region>
 
 # Multi-profile switching (alternative to --profile flag, useful for plugin integration)
 export AWS_MCP_PROXY_PROFILES="prod-readonly dev staging"
+
+# HTTP connection pool sizing (alternative to --max-connections / --max-keepalive-connections)
+# Raise the pool size when your MCP client issues many parallel aws_* tool calls per turn.
+export MCP_PROXY_MAX_CONNECTIONS=25
+export MCP_PROXY_MAX_KEEPALIVE_CONNECTIONS=5
 ```
 
 > **Note:** `AWS_MCP_PROXY_PROFILES` takes precedence over `--profile` / `AWS_PROFILE` when set.
+>
+> **Note:** `--max-connections` / `--max-keepalive-connections` take precedence over `MCP_PROXY_MAX_CONNECTIONS` / `MCP_PROXY_MAX_KEEPALIVE_CONNECTIONS` when explicitly passed. Defaults are `5` / `1`, preserving the previous behavior.
 
 ### Multi-account access
 
--- a/mcp_proxy_for_aws/cli.py
+++ b/mcp_proxy_for_aws/cli.py
@@ -17,7 +17,7 @@
 import argparse
 import os
 from mcp_proxy_for_aws import __version__
-from mcp_proxy_for_aws.utils import within_range
+from mcp_proxy_for_aws.utils import positive_int, within_range
 from typing import Any, Dict, Optional, Sequence
 
 
@@ -183,6 +183,24 @@ Examples:
         '--skip-auth',
         action='store_true',
         help='Skip request signing when AWS credentials are unavailable instead of failing',
+    )
+
+    parser.add_argument(
+        '--max-connections',
+        type=positive_int(1),
+        default=positive_int(1)(os.getenv('MCP_PROXY_MAX_CONNECTIONS', '5')),
+        help='Maximum number of concurrent HTTP connections in the pool used for all '
+        'proxied tool calls. Increase this when your MCP client fans out parallel '
+        'aws_* tool calls per turn (default: 5). '
+        'Falls back to MCP_PROXY_MAX_CONNECTIONS if not set.',
+    )
+
+    parser.add_argument(
+        '--max-keepalive-connections',
+        type=positive_int(0),
+        default=positive_int(0)(os.getenv('MCP_PROXY_MAX_KEEPALIVE_CONNECTIONS', '1')),
+        help='Maximum number of idle keep-alive connections to retain in the pool '
+        '(default: 1). Falls back to MCP_PROXY_MAX_KEEPALIVE_CONNECTIONS if not set.',
     )
 
     return parser.parse_args()
--- a/mcp_proxy_for_aws/middleware/profile_switcher.py
+++ b/mcp_proxy_for_aws/middleware/profile_switcher.py
@@ -66,6 +66,8 @@ class ProfileOverrideMiddleware(Middleware):
         endpoint: str,
         disable_telemetry: bool = False,
         skip_auth: bool = False,
+        max_connections: int = 5,
+        max_keepalive_connections: int = 1,
     ) -> None:
         """Initialize the middleware with connection and profile configuration."""
         super().__init__()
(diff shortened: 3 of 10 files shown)


valter-silva-au:
Friendly nudge on this one. Just rebased onto latest `main` (now even with base, single commit `c7ee76b`, no conflicts). The change is additive and default-preserving — no flags == today's `5/1` limits, asserted by test — and addresses #295. Full unit suite green (248 passed, 96% cov), ruff + pyright clean. @anasstahr you were engaged on the issue; would you or @krishmakochhar / @wzxxing be able to take a look when you have a moment? Happy to adjust anything.

valter-silva-au:
Still active from my side — please don't close.

Two things have changed since my last update: `main` has moved 18 commits ahead, so the branch now shows a conflict. I'll rebase and re-verify shortly, then re-request review.

The change itself is unchanged in intent and still relevant to #295: it makes the httpx connection pool size configurable via flags and env, additive and default-preserving (no flags == today's `5/1` limits, asserted by test).

@anasstahr you were engaged on the original issue — is this something the team still wants? If the answer is no, I'd rather close it than keep it on the stale treadmill. If yes, I'll get the rebase up.

PR #323, opened 2026-06-18
feat(profile): enable per-call aws_profile switching for eks-mcp tools

### LINKED ISSUE #322: Support profile switching for EKS tools
### Describe the feature (Security related? please follow <https://github.com/aws/mcp-proxy-for-aws/security/policy> to report them to AWS Security directly.)

This tool is cited in AWS's documentation as the preferred method for connecting to the eks-mcp service, yet it does not currently support profile switching.

### Use Case

I need this feature because my team supports many eks clusters across many accounts and we would like to be able to configure one mcp proxy for this instead of having a one-to-one ratio.

### Proposed Solution

After examining the code, it appears that all that is needed for this to work is to add the EKS tools to the list of tools that need the profile switching middleware.

### Other Information

_No response_

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

jared-schmidt-niceincontact:
### Changes

- Enables per-call AWS profile switching for the **eks-mcp** backend by adding its AWS/EKS/Kubernetes tools to `AUTH_REQUIRING_TOOLS` in `ProfileOverrideMiddleware`.

`ProfileOverrideMiddleware` is already service-agnostic — tool listing, invocation, and per-profile routing all gate solely on membership in `AUTH_REQUIRING_TOOLS` and operate on `tool.name`. So the change is just extending that set:
- Added the **16** eks-mcp tools that actually call AWS/EKS/Kubernetes. These arrive as **bare names** (e.g. `manage_k8s_resource`, `apply_yaml`), with no `aws___` gateway prefix, so the bare strings are matched.
- Intentionally **excluded** the 4 pure documentation/guidance/manifest-generation tools (`search_eks_documentation`, `search_eks_troubleshooting_guide`, `get_eks_metrics_guidance`, `generate_app_manifest`) since they make no AWS call — mirroring how the AWS-API allowlist already omits doc/metadata tools like `search_documentation`/`list_regions`.
No routing, prefix, or transport code changes were needed.

## User Experience

The proxy is the documented way to connect to eks-mcp, but per-call profile switching only worked for the AWS API server's tools. Teams managing many EKS clusters across accounts need a single proxy that can target a different account per call (e.g. read pods in `dev` vs `prod`) instead of running one proxy per account.

## Checklist

* [x] Changes are documented

* [ ] Yes
* [x] No

- [x] Workflow syntax validated

### Testing details

- Added unit tests: every one of the 16 eks tools gets `aws_profile` injected on `list_tools`; the 4 doc tools do not; and an eks tool called with a non-default `aws_profile` routes through a dedicated per-profile client with the argument stripped.
- `uv run pytest tests/unit/test_profile_switcher.py` — all pass (38).
- `ruff check`, `ruff format --check`, and `pyright` clean.

### Other notes

Closes #322

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/middleware/profile_switcher.py
+++ b/mcp_proxy_for_aws/middleware/profile_switcher.py
@@ -83,11 +83,29 @@ class ProfileOverrideMiddleware(Middleware):
 
     # Tools that require AWS authentication and support profile switching
     AUTH_REQUIRING_TOOLS = {
+        # AWS API MCP server
         'aws___call_aws',
         'aws___run_script',
         'aws___get_presigned_url',
         'aws___get_tasks',
         'aws___suggest_aws_commands',
+        # EKS MCP server (tools that call AWS/EKS/Kubernetes; doc/guidance tools omitted)
+        'add_inline_policy',
+        'apply_yaml',
+        'describe_eks_resource',
+        'get_cloudwatch_logs',
+        'get_cloudwatch_metrics',
+        'get_eks_insights',
+        'get_eks_vpc_config',
+        'get_k8s_events',
+        'get_pod_logs',
+        'get_policies_for_role',
+        'list_api_versions',
+        'list_eks_resources',
+        'list_k8s_resources',
+        'manage_eks_stacks',
+        'manage_k8s_resource',
+        'read_k8s_resource',
     }
 
     # ── tool listing ────────────────────────────────────────────────
--- a/README.md
+++ b/README.md
@@ -157,6 +157,17 @@ AWS_MCP_PROXY_PROFILES="prod-readonly dev staging" mcp-proxy-for-aws https://aws
 - Invalid profiles are rejected with an error listing allowed values
 - Each non-default profile gets its own dedicated connection to the backend
 
+This works for both the AWS API MCP server and the EKS MCP server. For example, pointing the
+proxy at the eks-mcp endpoint lets an agent read or manage Kubernetes resources across accounts:
+
+```bash
+mcp-proxy-for-aws https://eks-mcp.us-east-1.api.aws/mcp --profile prod-readonly dev
+```
+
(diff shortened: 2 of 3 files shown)


mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -83,11 +83,29 @@ def __init__(
 
     # Tools that require AWS authentication and support profile switching
     AUTH_REQUIRING_TOOLS = {
+        # AWS API MCP server
         'aws___call_aws',
         'aws___run_script',
         'aws___get_presigned_url',
         'aws___get_tasks',
         'aws___suggest_aws_commands',
+        # EKS MCP server (tools that call AWS/EKS/Kubernetes; doc/guidance tools omitted)
```
patrick-yu-amzn, line 92: Should we have some sort of namespacing here to avoid accidentally allowlisting tools from other mcp servers?

mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -83,11 +83,29 @@ def __init__(
...
+        'add_inline_policy',
```
patrick-yu-amzn, line 93: The example given in README.md actually highlighted my two concerns.

First, assuming the customer wants to use a dev profile that can read/write along with a prod profile that can only read, the proxy itself should not be configured with the `--read-only` flag that hides all write tools. However, there's no indication to the model that write tools with the prod profile won't work. Hopefully the models would be adaptive enough and figure it out eventually.

Secondly, we're relying on the customer to ensure the prod profile doesn't have write permissions. Models are all too eager to `oh this fix worked fine in dev cluster, lemme apply it to the prod cluster as well`, which doesn't always end well. Offering profile switching that mixes read/write can be risky.

PR #350, opened 2026-07-07
fix: enforce read-only mode on tools/call requests

anasstahr:
### Changes

The `--read-only` flag previously only filtered tools from `tools/list` responses. A client that knew the tool name could still invoke it directly via `tools/call`. This adds an `on_call_tool` hook to `ToolFilteringMiddleware` that rejects calls to tools not in the allowed set when `--read-only` is active.

### User experience

**Before:** A client could bypass `--read-only` by directly calling a write tool by name (e.g. `aws___call_aws`), even though it was hidden from `tools/list`.

**After:** Any `tools/call` request targeting a tool that was filtered out in read-only mode is rejected with a `ToolError`.

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

Tested manually by piping JSON-RPC messages over stdio with `--read-only` and confirming that `tools/call` to a write tool is now rejected.

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/middleware/tool_filter.py
+++ b/mcp_proxy_for_aws/middleware/tool_filter.py
@@ -13,9 +13,11 @@
 # limitations under the License.
 
 import logging
+import mcp.types as mt
 from collections.abc import Awaitable, Callable
-from fastmcp.server.middleware import Middleware, MiddlewareContext
-from fastmcp.tools import Tool
+from fastmcp.exceptions import ToolError
+from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
+from fastmcp.tools import Tool, ToolResult
 from typing import Sequence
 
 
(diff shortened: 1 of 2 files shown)


mcp_proxy_for_aws/middleware/tool_filter.py
```diff
@@ -55,4 +58,20 @@ async def on_list_tools(
...
+    async def on_call_tool(
+        self,
+        context: MiddlewareContext[mt.CallToolRequestParams],
+        call_next: CallNext[mt.CallToolRequestParams, ToolResult],
+    ) -> ToolResult:
+        """Reject calls to tools that are not allowed in read-only mode."""
+        if not self.read_only:
+            return await call_next(context)
+
+        tool_name = context.message.name
```
arnewouters: I would prefer doing something similar as the example in the docs instead of storing a list: https://gofastmcp.com/servers/middleware#filtering-lists

We can just check the annotation on the tools again

mcp_proxy_for_aws/middleware/tool_filter.py
```diff
@@ -56,3 +58,21 @@ async def on_list_tools(
             filtered_tools.append(tool)
 
         return filtered_tools
+
+    async def on_call_tool(
```
arnewouters, line 62: Please add a unit test as well

arnewouters (changes requested):

arnewouters (approved):

PR #357, opened 2026-07-17
fix: pass resolved region into boto session

### LINKED ISSUE #342: NoRegionError during local credential refresh when active profile has no region (toolkit#63)
## Summary

`mcp-proxy-for-aws` raises `botocore.exceptions.NoRegionError: You must specify a region` during local credential refresh when:

- The endpoint URL or `--region` flag already supplies a valid signing region (so the proxy itself starts up "correctly"), AND
- The active AWS profile has no persisted `region = ...` line, AND
- `AWS_REGION` / `AWS_DEFAULT_REGION` are not set in the proxy process environment.

This is common with AWS CLI v2 `aws login` / SSO-style credentials, where `~/.aws/config` ends up like:

```ini
[default]
login_session = arn:aws:iam::<acct>:user/<redacted>
```

— no `region = ...` line. From the user's perspective everything is configured (URL has the region, `--metadata AWS_REGION=...` set), so the failure surfaces as an opaque `JSON-RPC error: -32602: Invalid request parameters("")` to the MCP client. `aws sso login` does not help because the missing setting is in the local boto session, not the credentials.

Originally reported in [aws/agent-toolkit-for-aws#63](https://github.com/aws/agent-toolkit-for-aws/issues/63), where the immediate workaround turned out to be:

```json
"env": { "AWS_REGION": "...", "AWS_DEFAULT_REGION": "..." }
```

…but that workaround duplicates information the proxy already has.

## Why this happens

The proxy resolves its **signing** region through a layered chain — `--region` flag → URL parsing → `AWS_REGION` env (`utils.py::determine_aws_region`). That chain works.

The **local botocore session** does not get any of those:

```python
# mcp_proxy_for_aws/sigv4_helper.py:158
session = boto3.Session(profile_name=profile) if profile else boto3.Session()
```

No `region_name` is passed in. botocore then resolves region only via env vars and the active profile's `region` line. If neither has one, any internal boto client created for credential resolution (e.g. the `signin` client during SSO/`aws login` refresh) raises `NoRegionError` before the proxy ever talks to the MCP server.

## Repro

1. Configure an `aws login` / SSO profile with no persisted `region` line:
   ```ini
   [default]
   login_session = arn:aws:iam::<acct>:user/<u>
   ```
2. Ensure `AWS_REGION` / `AWS_DEFAULT_REGION` are not set in the shell that launches the MCP client.
3. Configure any MCP client (Codex/Kiro/Cursor) with:
   ```json
   {
     "command": "uvx",
     "args": [
       "mcp-proxy-for-aws@latest",
       "--metadata", "AWS_REGION=ap-northeast-1"
     ]
   }
   ```
4. Start the MCP client. Initialize fails with `-32602 Invalid request parameters("")`.
5. `mcp-proxy-for-aws --log-level DEBUG` shows `botocore.exceptions.NoRegionError: You must specify a region.` during credential refresh.

## Proposed fix

Pass the proxy's already-resolved region (the same value used for SigV4 signing) into the local boto session so botocore's internal clients have a default region available during credential refresh:

```python
# mcp_proxy_for_aws/sigv4_helper.py
def create_aws_session(profile: Optional[str] = None, region: Optional[str] = None) -> boto3.Session:
    kwargs: Dict[str, Any] = {}
    if profile:
        kwargs["profile_name"] = profile
    if region:
        kwargs["region_name"] = region
    return boto3.Session(**kwargs)
```

…and thread the already-determined region through `create_sigv4_client` → `_sign_request_hook` → `create_aws_session`. The proxy already has this value at `server.py::run_proxy` line 60 (`region = determine_aws_region(args.endpoint, args.region)`).

This is a no-op for users who *do* set `AWS_REGION` or have a profile region, and fixes the failure for users who don't.

### Why not just document the env-block workaround

That's what aws/agent-toolkit-for-aws#63 originally proposed, and we opened [aws/agent-toolkit-for-aws#120](https://github.com/aws/agent-toolkit-for-aws/pull/120) to do it — then closed it in favor of fixing it here. The proxy already infers a signing region; teaching every client config (Kiro/Codex/Cursor/Claude Code) to re-specify the same region in an env block is duplicate configuration and doesn't help users who hit this in already-deployed MCP setups.

## Related but distinct

- #321 / #320 cover the `login_session` refresh **boundary** behavior (refresh fails mid-session). This issue is about the **startup-time** `NoRegionError` when the local profile has no region. The two failure modes can coexist but have different root causes.
- #304 covers `LoginRefreshRequired` being masked as `-32602`. Same surface error from the MCP client perspective, different underlying cause.

## Environment

- mcp-proxy-for-aws: 1.6.x and main
- AWS CLI v2 `aws login` / IAM Identity Center SSO
- Any MCP client that proxies the `-32602` initialize error back to the user (Codex, Kiro, Cursor, Claude Code)

miachillgood:
## Summary
- pass the already-resolved proxy region into `boto3.Session` creation
- let local botocore credential refresh reuse the proxy region instead of requiring separate env configuration
- update unit coverage for the region-aware session creation path

## Testing
- `git diff --check`
- local Python test dependencies were not available in this environment, so pytest could not be run here

Closes #342.

### DIFF
--- a/mcp_proxy_for_aws/sigv4_helper.py
+++ b/mcp_proxy_for_aws/sigv4_helper.py
@@ -154,11 +154,12 @@ class SigV4HTTPXAuth(httpx.Auth):
         yield request
 
 
-def create_aws_session(profile: Optional[str] = None) -> boto3.Session:
-    """Create an AWS session with optional profile.
+def create_aws_session(profile: Optional[str] = None, region: Optional[str] = None) -> boto3.Session:
+    """Create an AWS session with optional profile and region.
 
     Args:
         profile: AWS profile to use (optional)
+        region: AWS region to bind to the boto session (optional)
 
     Returns:
         boto3.Session instance
@@ -167,7 +168,12 @@ def create_aws_session(profile: Optional[str] = None) -> boto3.Session:
         ValueError: If session creation fails
     """
     try:
-        session = boto3.Session(profile_name=profile) if profile else boto3.Session()
+        session_kwargs: Dict[str, Any] = {}
+        if profile:
+            session_kwargs['profile_name'] = profile
+        if region:
+            session_kwargs['region_name'] = region
+        session = boto3.Session(**session_kwargs)
     except Exception as e:
         raise ValueError(f"Failed to create AWS session with profile '{profile}': {e}")
 
@@ -316,7 +322,7 @@ async def _sign_request_hook(
 
     # Always read fresh credentials from disk so account switches
     # and credential refreshes take effect immediately.
-    session = create_aws_session(profile)
+    session = create_aws_session(profile, region)
     credentials = session.get_credentials()
 
     if credentials is None:
--- a/tests/unit/test_hooks.py
+++ b/tests/unit/test_hooks.py
@@ -379,7 +379,7 @@ class TestSignRequestHook:
 
         await _sign_request_hook('us-east-1', 'execute-api', 'my-profile', False, request)
 
-        mock_create_session.assert_called_once_with('my-profile')
+        mock_create_session.assert_called_once_with('my-profile', 'us-east-1')
 
     @pytest.mark.asyncio
     async def test_sign_request_hook_signs_request(self, mock_create_session):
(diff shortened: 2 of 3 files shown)


anasstahr:
Thanks for the contribution! We'd prefer not to override boto3's built-in region resolution and instead let users configure their environment correctly, closing the PR

PR #363, opened 2026-07-20
fix: stale profile refresh when using multi-profiles

### LINKED ISSUE #320: Non-default profile clients remain stale after credential refresh while default profile recovers
## Environment
- `mcp-proxy-for-aws` 1.6.0 (also reproduces on main as of 2026-06-18)
- AWS CLI v2 with `aws login` (`login_session` in `~/.aws/config`)
- Windows 11, Python 3.10.18
- `--profile stg dev` (stg = default, dev = non-default)

## Expected behavior

After credentials are refreshed on disk (e.g. via `aws sts get-caller-identity`), all profiles — both default and non-default — should recover and serve MCP requests successfully within the same proxy process.

## Actual behavior

After running:
```
aws sts get-caller-identity --profile stg
aws sts get-caller-identity --profile dev
```

- **stg (default profile)**: ✅ recovers immediately
- **dev (non-default profile)**: ❌ remains broken until process restart

The non-default profile continues to fail with:
```
Tool call failed using profile 'dev'. The request could not be completed with the specified profile.
```

This asymmetry is not dev/stg-specific. When profile order is reversed (`--profile dev stg`):

- **dev (default)**: ✅ recovers
- **stg (non-default)**: ❌ remains broken

The failure follows the **default/non-default position**, not the profile name.

## Reproduction steps

1. `aws login --profile stg && aws login --profile dev`
2. Start proxy: `mcp-proxy-for-aws ... --profile stg dev`
3. Confirm both profiles work via MCP
4. Wait until the credential refresh/expiry boundary (~10–15 min), where both profiles begin failing in my environment
5. Run `aws sts get-caller-identity --profile stg && aws sts get-caller-identity --profile dev`
6. Test both profiles via MCP → stg (default) recovers, dev (non-default) does not

## Suspected cause

`ProfileOverrideMiddleware._get_profile_client()` caches `fastmcp.Client` instances in `_profile_clients` on first use and never invalidates them:

```python
async def _get_profile_client(self, profile: str) -> Client:
    async with self._lock:
        if profile not in self._profile_clients:
            transport = create_transport_with_sigv4(...)
            client = Client(transport=transport)
            await client.__aenter__()
            self._profile_clients[profile] = client  # cached for the proxy process lifetime
        return self._profile_clients[profile]  # no reconnect on failure
```

When a credential-related failure occurs, the cached client's underlying MCP session appears to become unusable. Unlike `AWSMCPProxyClient` (used for the default profile), which has self-healing reconnect logic in `_connect()`:

```python
except RuntimeError as e:
    await self._disconnect(force=True)
    return await self._connect(retry + 1)  # reconnects with fresh credentials
```

`_get_profile_client()` has no equivalent recovery path. Refreshing credentials on disk with `aws sts get-caller-identity` does not recover the non-default profile within the same proxy process — the stale cached client continues to be returned.

## Suggested fix

On auth/session-related failures, invalidate the cached per-profile client and either retry once with a fresh client or allow the next request to reconnect. The stale client should also be closed before being discarded:

```python
except Exception:
    async with self._lock:
        stale = self._profile_clients.pop(profile, None)
    if stale is not None:
        try:
            await stale.__aexit__(None, None, None)
        except Exception:
            logger.debug('Failed to close stale profile client', exc_info=True)
    raise
```

With this change, after `aws sts get-caller-identity --profile stg && aws sts get-caller-identity --profile dev`, both profiles recover immediately within the same proxy process — verified locally on 2026-06-18.

arnewouters:
### Changes

Fixes #320. Per-profile MCP clients were cached forever with no invalidation, so a non-default profile whose session died during the credential-expiry window stayed broken until proxy restart (the default profile self-heals via `AWSMCPProxyClient._connect()`).

On a session-level failure, the middleware now removes and closes the cached client (identity-safe, only if it's still the failed instance) so the next call reconnects with fresh credentials. No automatic retry; tool-level errors (`is_error`) don't invalidate. `disconnect_profile_clients()` now snapshots the cache under the lock.

### User experience

Before: after credentials expire and are refreshed on disk, non-default profiles keep failing with "Tool call failed using profile ..." until the proxy is restarted.

After: the next call on that profile reconnects and succeeds.

## Checklist

* [ ] Changes are documented

* [ ] Yes
* [x] No

Five new unit tests: invalidation without replay, fresh-client reconnect on next call, cleanup failure not masking the original error, `is_error` results leaving the session cached, and a stale-client/replacement identity race.

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/middleware/profile_switcher.py
+++ b/mcp_proxy_for_aws/middleware/profile_switcher.py
@@ -188,14 +188,34 @@ class ProfileOverrideMiddleware(Middleware):
                 self._profile_clients[profile] = client
             return self._profile_clients[profile]
 
+    async def _invalidate_profile_client(self, profile: str, failed_client: Client) -> None:
+        """Remove and disconnect a failed client if it is still cached."""
+        stale_client: Client | None = None
+        async with self._lock:
+            if self._profile_clients.get(profile) is failed_client:
+                stale_client = self._profile_clients.pop(profile)
+
+        if stale_client is not None:
+            try:
+                await stale_client.__aexit__(None, None, None)
+            except Exception:
+                logger.warning(
+                    'Failed to disconnect invalidated profile client %s',
+                    profile,
+                    exc_info=True,
+                )
+
     async def disconnect_profile_clients(self) -> None:
         """Disconnect all per-profile clients. Call during server shutdown."""
-        for profile, client in self._profile_clients.items():
+        async with self._lock:
+            profile_clients = list(self._profile_clients.items())
+            self._profile_clients.clear()
+
+        for profile, client in profile_clients:
             try:
                 await client.__aexit__(None, None, None)
             except Exception:
                 logger.exception('Failed to disconnect profile client %s', profile)
-        self._profile_clients.clear()
 
     async def _call_with_profile(
         self,
@@ -252,6 +272,7 @@ class ProfileOverrideMiddleware(Middleware):
             raise
         except Exception as e:
             logger.exception('Error calling tool via profile %s', profile)
+            await self._invalidate_profile_client(profile, client)
             raise ToolError(
                 f'Tool call failed using profile {profile!r}. '
                 'The request could not be completed with the specified profile.'
--- a/tests/unit/test_profile_switcher.py
+++ b/tests/unit/test_profile_switcher.py
@@ -271,6 +271,155 @@ class TestPerCallProfileOverride:
         assert 'backend error' not in str(exc_info.value)
 
 
+class TestProfileClientInvalidation:
+    """Tests for invalidating failed per-profile clients."""
+
+    @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


mcp_proxy_for_aws/middleware/profile_switcher.py
```diff
@@ -252,6 +272,7 @@ async def _call_with_profile(
             raise
         except Exception as e:
             logger.exception('Error calling tool via profile %s', profile)
+            await self._invalidate_profile_client(profile, client)
```
anasstahr, line 275: wont this mean that we will get a failed request, invalidate the client and then the next request will actually work? not 100% sure, but for default we retry the request immediately when it fails because of credentials, cant we do something similar

arnewouters: I'm not sure what you mean? What's the point of immediately retrying on a failed request, we need the user to refresh the credentials first, no?

anasstahr (approved):

PR #381, opened 2026-08-05
feat: two-vend packaging + build-time dependency pinning

meghagoyal04:
## What
Introduces **two-vend packaging** for MCP Proxy for AWS from one codebase:

- **`mcp-proxy-for-aws`** — a metadata-only *pinned* wrapper for CLI/`uvx` users. Its exact `==` runtime closure (plus `mcp-proxy-for-aws-lib==<v>`) is **derived from `uv.lock` at build time** via a Hatchling custom metadata hook (`packages/proxy/hatch_build.py`) and baked into the published wheel metadata.
- **`mcp-proxy-for-aws-lib`** — the existing package, renamed, keeping the code and **loose** dependency ranges for library/programmatic users. Import name `mcp_proxy_for_aws` is unchanged.

## Why
`uvx mcp-proxy-for-aws@<v>` today pins only our own version; the transitive runtime closure resolves to newest on each cold install, so a newer **compromised release** of any dependency is auto-adopted. The pinned wrapper version-pins the whole runtime tree for CLI users, while library users keep co-resolvable loose ranges.

> Scope: this is **version-pinning, not hash-verification** — it prevents auto-adoption of a newer compromised release; it does not cryptographically verify the installed bits.

## How
- uv **workspace** with the wrapper at `packages/proxy/`, deliberately **excluded** from the workspace so `uv lock` doesn't build its dynamic metadata from the lock it's writing.
- The hook derives pins via `uv export --package mcp-proxy-for-aws-lib --no-dev` (the library's runtime closure only — **not** the full lock, so dev-groups and example-client trees are excluded), markers preserved.
- From-sdist builds read a baked `_runtime_pins.txt` (generated at sdist build; gitignored); wheels carry static metadata so end users never run the hook.
- **Pre-release guard:** the build fails if any derived pin is a pre-release (uninstallable without `--prerelease=allow`).
- Dropped an invalid `[tool.uv] exclude-newer = "P3D"` and re-locked so the closure is all-stable.

## Testing
- Hook + packaging unit tests under `packages/proxy/tests/`.
- Local `uv build` → wrapper wheel is metadata-only, `Requires-Dist` = 88 pins, **no pre-releases**, dev/example deps absent.
- Flag-free end-to-end: `uvx`-style install of the wrapper resolves the frozen tree and `mcp-proxy-for-aws --help` runs.


packages/proxy/hatch_build.py
```diff
@@ -0,0 +1,204 @@
...
+    Args:
+        project_root: The wrapper's project root (the directory holding its pyproject).
+
+    Returns:
+        The exact runtime closure of the library, with environment markers preserved.
+    """
+    baked = project_root / BAKED_PINS_FILENAME
+    if baked.is_file():
+        return _reject_prereleases(_read_baked_pins(baked))
+    # In the repo the wrapper lives at <workspace>/packages/proxy.
```
anasstahr: maybe better to have a check to confirm it lives there/throw an error if it doesnt? not sure what error we get now

packages/proxy/tests/test_hook.py
```diff
@@ -0,0 +1,287 @@
...
+            {},
+            {},
+            None,
+            str(tmp_path),
+            'sdist',  # type: ignore[arg-type]
+        )
+        build_data: dict = {}
+        try:
+            hook.initialize('standard', build_data)
+            pins_path = PROJECT_ROOT / BAKED_PINS_FILENAME
```
anasstahr: shouldnt this be tmp_path?

pyproject.toml
```diff
@@ -1,14 +1,17 @@
-[tool.uv]
-exclude-newer = "P3D"
```
anasstahr, line 2: shouldnt we keep this?

pyproject.toml
```diff
@@ -50,6 +53,11 @@ changelog = "https://github.com/aws/aws-mcp-proxy/blob/main/CHANGELOG.md"
 [dependency-groups]
 dev = [
     "commitizen>=4.2.2",
+    # Build backend, needed in the dev environment to unit-test packages/proxy's
+    # build hooks (which import hatchling's plugin interfaces).
+    "hatchling>=1.27.0",
+    # tomllib backport, for tests that read pyproject.toml on Python 3.10.
+    "tomli>=2.0.0; python_full_version < '3.11'",
```
arnewouters, line 76: Why is this Python 3.10 only?

packages/proxy/LICENSE
```diff
@@ -0,0 +1,175 @@
+
```
arnewouters, line 1: Can we symlink these files instead?

arnewouters (commented):
Do we not need to update the release workflow?

Let's also test with https://test.pypi.org to see that everything works for both releases

packages/proxy/hatch_build.py
meghagoyal04: Fixed in the new commit. Today a missing directory raised FileNotFoundError, which got misreported as "uv was not found on PATH". We now verify uv.lock is present before invoking uv, and the error names both inputs.

packages/proxy/tests/test_hook.py
meghagoyal04: tmp_path was already there as the 5th arg. But you were right it shouldn't touch the tree, and  chasing that found a bug - a leftover _runtime_pins.txt outranked uv.lock, building a wheel with 2 pins instead of 88, silently. Both fixed

pyproject.toml
meghagoyal04: Cooldown moved to Dependabot

pyproject.toml
meghagoyal04: tomllib is stdlib from 3.11, so tomli is only needed on 3.10: the marker keeps the backport off newer interpreters

packages/proxy/LICENSE
meghagoyal04: Tried it, and it breaks the sdist -> wheel path. Git stores the symlink

anasstahr (approved):

rshevchuk-git:
Wouldn't this break the pip upgrade flow? Both distributions install the same folder, and pip's uninstall deletes by file path with no awareness that another distribution now depends on those paths. If a user decides to upgrade via `pip install --upgrade`, it will break their `mcp-proxy-for-aws` CLI. We need to test this

meghagoyal04:
> Wouldn't this break the pip upgrade flow? Both distributions install the same folder, and pip's uninstall deletes by file path with no awareness that another distribution now depends on those paths. If a user decides to upgrade via `pip install --upgrade`, it will break their `mcp-proxy-for-aws` CLI. We need to test this

Confirmed this failure mode and it's silent: upgrading from the published 1.6.3 to the new  pair deletes mcp_proxy_for_aws/, so the CLI dies with ModuleNotFoundError while pip check still reports nothing wrong. pip writes the module for  -lib, then deletes it while uninstalling old 1.6.3, whose RECORD still  lists those paths.
  
One-time transition (the new pair itself doesn't share paths) and pip-only  -- uvx and uv pip --upgrade both handle it. Recovery is `pip install --force-reinstall --no-deps mcp-proxy-for-aws-lib`. Do we takethe one-time break with a migration note, or flip which distribution is the new one so nothing changes file ownership?

rshevchuk-git:
I think you meant -force-reinstall `mcp-proxy-for-aws` as reinstalling -lib silently defeats the pinning. The issue is that it's hard for a user to derive what's going wrong from the ModuleNotFoundError error...

arnewouters:
We shipped #392 instead

PR #392, opened 2026-08-24
feat: ship a version-pinned CLI distribution (mcp-proxy-for-aws-cli)

meghagoyal04:
Publish two distributions from one codebase so CLI users get a frozen dependency tree without constraining library users:

- `mcp-proxy-for-aws` keeps the code and loose, co-resolvable dependency ranges for programmatic/library use. The import name `mcp_proxy_for_aws` and the `mcp-proxy-for-aws` console script are unchanged.
- `mcp-proxy-for-aws-cli` is a new, metadata-only wrapper for CLI/`uvx` use. Its exact runtime closure (plus `mcp-proxy-for-aws==<version>`) is derived from uv.lock at build time by a Hatchling metadata hook and baked into the published wheel, so a fresh install gets a frozen tree and cannot silently adopt a newer release of a transitive dependency.

The wrapper is excluded from the uv workspace so `uv lock` never builds its dynamic metadata from the lock it is writing. A build-time guard fails the release if the closure contains a pre-release (which uvx/pip users cannot install without --prerelease=allow). Publish workflows build the wrapper by path and publish the code package before the wrapper that pins it; the Dockerfile installs mcp-proxy-for-aws-cli; the READMEs document which distribution to use. python.yml also gains a merge_group trigger so the required Build check runs for queued merges.

This is version-pinning, not hash-verification: it prevents auto-adoption of a newer compromised release; it does not cryptographically verify installed artifacts.

### Changes

Publishes two distributions from this one codebase so CLI users get a frozen dependency tree without constraining library users:
- **`mcp-proxy-for-aws`** — unchanged: keeps the code and its loose, co-resolvable dependency ranges. The import name `mcp_proxy_for_aws` and the `mcp-proxy-for-aws` console script are the same, so existing installs and programmatic users are unaffected.
- **`mcp-proxy-for-aws-cli`** — new, metadata-only wrapper for CLI/`uvx` use. Its exact `==` runtime closure (plus `mcp-proxy-for-aws==<version>`) is derived from `uv.lock` at build time by a Hatchling metadata hook and baked into the  published wheel, so a fresh install gets a frozen tree and cannot silently adopt a newer release of a transitive dependency
 
Implementation notes:
-  The wrapper is excluded from the uv workspace so `uv lock` never builds its dynamic metadata from the lock file it is writing.
- A build-time guard fails the release if the pinned closure contains a pre-release (uninstallable via `uvx`/`pip` without `--prerelease=allow`).
- Publish workflows build the wrapper by path and publish the code package **before** the wrapper that pins it by exact version; the Dockerfile installs `mcp-proxy-for-aws-cli`.
- `python.yml` gains a `merge_group` trigger so the required Build check runs for merge-queue entries.
- READMEs document which distribution to install for CLI vs. library use

### User experience

**Before**
`uvx mcp-proxy-for-aws@<version>` pins only the proxy's own version. Its transitive dependencies are loose ranges, so every cold install resolves the whole tree to whatever is newest on PyPI — a newer, possibly compromised release
  of a dependency is adopted automatically.
  
**After**
 - **CLI / `uvx` users** install the pinned distribution and get a frozen tree:
   uvx mcp-proxy-for-aws-cli@<version> <SigV4 MCP endpoint URL>
  "args": ["mcp-proxy-for-aws-cli@<version>", "https://aws-mcp.us-east-1.api.aws/mcp"]
  
- Library users are unchanged:  same install, same import, same loose ranges so it co-resolves with their own dependencies:
  pip install mcp-proxy-for-aws
  from mcp_proxy_for_aws.client import aws_iam_streamablehttp_client
  
- Existing mcp-proxy-for-aws installs keep working and upgrade normally. That distribution still owns the code and the mcp-proxy-for-aws command, so a  pip install --upgrade is an same-distribution upgrade 
  

The README points CLI/uvx users at mcp-proxy-for-aws-cli and library users at mcp-proxy-for-aws, so new setups and reconfigurations pick the right one.

## Checklist

* [X] Changes are documented

* [ ] Yes
* [X] No

## Acknowledgment

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,6 +28,13 @@ updates:
       day: "monday"
       time: "07:00"
       timezone: "Etc/UTC"
+    # Supply-chain cooldown: don't adopt a release until it has been public for a few days,
+    # so a compromised upload has time to be yanked before it reaches our lock (and, via
+    # the build-time pin hook, the published mcp-proxy-for-aws wrapper). This is the right
+    # home for the cooldown; `[tool.uv] exclude-newer` only accepts a fixed date, not a
+    # rolling window, so it cannot express this.
+    cooldown:
+      default-days: 3
     commit-message:
       prefix: "chore(deps): update uv"
     groups:
@@ -60,6 +67,9 @@ updates:
       day: "monday"
       time: "07:00"
       timezone: "Etc/UTC"
+    # See the cooldown note on the `uv` ecosystem above.
+    cooldown:
+      default-days: 3
     commit-message:
       prefix: "chore(deps): update pip"
     groups:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,9 +30,10 @@ updates:
       timezone: "Etc/UTC"
     # Supply-chain cooldown: don't adopt a release until it has been public for a few days,
     # so a compromised upload has time to be yanked before it reaches our lock (and, via
-    # the build-time pin hook, the published mcp-proxy-for-aws wrapper). This is the right
-    # home for the cooldown; `[tool.uv] exclude-newer` only accepts a fixed date, not a
-    # rolling window, so it cannot express this.
+    # the build-time pin hook, the published mcp-proxy-for-aws wrapper). We keep the
+    # cooldown here rather than in uv's `exclude-newer` so it is independent of the uv
+    # version resolving the lock: recent uv can express a rolling window via an
+    # `exclude-newer` duration, but older uv rejects durations and silently drops the setting.
     cooldown:
       default-days: 3
     commit-message:
--- a/.github/workflows/pypi-publish-on-release.yml
+++ b/.github/workflows/pypi-publish-on-release.yml
@@ -68,8 +68,19 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@e4db8464a088ece1b920f60402e813ea4de65b8f # v4
 
-      - name: Build distribution packages
+      # Two distributions are released from this one commit: the library (which holds the
+      # code) and the metadata-only `mcp-proxy-for-aws` wrapper that pins the library's
+      # exact runtime closure. The wrapper is deliberately outside the uv workspace -- its
+      # dependencies are derived from uv.lock at build time -- so neither `uv build` nor
+      # `uv build --all-packages` can see it. It needs its own invocation.
+      - name: Build the library distribution
         run: uv build
+
+      # `-o dist` is required, not cosmetic: `uv build <path>` defaults to <path>/dist, so
+      # without it the wrapper lands in packages/proxy/dist and every later step here
+      # (attestations, upload, publish) silently skips it -- they all glob ./dist.
+      - name: Build the pinned wrapper distribution
+        run: uv build packages/proxy -o dist
 
       - name: Generate SLSA provenance attestations
         uses: actions/attest-build-provenance@4d101475d8b20a2381f78447822ac1eab6504dd8 # v4.2.2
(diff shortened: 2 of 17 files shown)


.github/workflows/pypi-publish-on-release.yml
```diff
@@ -68,9 +68,20 @@ jobs:
...
         uses: astral-sh/setup-uv@e4db8464a088ece1b920f60402e813ea4de65b8f # v4
 
-      - name: Build distribution packages
+      # Two distributions are released from this one commit: the library (which holds the
+      # code) and the metadata-only `mcp-proxy-for-aws` wrapper that pins the library's
+      # exact runtime closure. The wrapper is deliberately outside the uv workspace -- its
+      # dependencies are derived from uv.lock at build time -- so neither `uv build` nor
+      # `uv build --all-packages` can see it. It needs its own invocation.
+      - name: Build the library distribution
         run: uv build
```
arnewouters, line 77: Will the `packages/proxy` folder be bundled here?

README.md
```diff
@@ -53,12 +57,14 @@ The MCP Proxy serves as a lightweight, client-side bridge between MCP clients (A
...
+
 ```bash
-# Run the server
-uvx mcp-proxy-for-aws@1.6.0 <SigV4 MCP endpoint URL>
+# Run the server (pinned CLI distribution)
+uvx mcp-proxy-for-aws-cli@1.7.0 <SigV4 MCP endpoint URL>
 ```
 
-**Note:** It is recommended to pin to a specific version (e.g., `@1.6.0`) to ensure reproducible behavior. Using `@latest` may pull in breaking changes. Check [PyPI](https://pypi.org/project/mcp-proxy-for-aws/) for the latest stable version.
+**Note:** It is recommended to pin to a specific version (e.g., `@1.7.0`) to ensure reproducible behavior. Using `@latest` may pull in breaking changes. Check [PyPI](https://pypi.org/project/mcp-proxy-for-aws-cli/) for the latest stable version.
```
arnewouters: Can we remove this and put `@latest` again everywhere?

.github/workflows/pypi-publish-on-release.yml
meghagoyal04: Good catch. it's in the sdist but not the wheel. Added [tool.hatch.build.targets.sdist] exclude = ["packages", ".agents"] to the root pyproject; verified packages/proxy is now gone from the code sdist and the wheel is still module-only.

README.md
meghagoyal04: Updated

pyproject.toml
```diff
@@ -1,11 +1,30 @@
-[tool.uv]
-exclude-newer = "P3D"
-prerelease = "allow"
+# NOTE: there is deliberately no `[tool.uv]` table here.
+#
+# It previously held `exclude-newer = "P3D"` and `prerelease = "allow"`, and both were
+# inert: `exclude-newer` only accepts a fixed date, so `"P3D"` failed to parse and uv
+# discarded the *whole* table along with it -- including the `prerelease` setting.
```
arnewouters: I'm confused by this, `P3D` is a valid value https://docs.astral.sh/uv/reference/settings/#dependency-metadata

.github/dependabot.yml
```diff
@@ -28,6 +28,13 @@ updates:
       day: "monday"
       time: "07:00"
       timezone: "Etc/UTC"
+    # Supply-chain cooldown: don't adopt a release until it has been public for a few days,
+    # so a compromised upload has time to be yanked before it reaches our lock (and, via
+    # the build-time pin hook, the published mcp-proxy-for-aws wrapper). This is the right
+    # home for the cooldown; `[tool.uv] exclude-newer` only accepts a fixed date, not a
+    # rolling window, so it cannot express this.
```
arnewouters: Not sure what the interaction is between `exclude-newer` from uv and dependabot but I think the comment is wrong, see my other comment

pyproject.toml
meghagoyal04: oh this was changed in this version of uv

.github/dependabot.yml
meghagoyal04: I kept the cooldown in Dependabot rather than switching to [tool.uv] exclude-newer = "P3D", for one reason: the repo pins no uv version, so on any contributor's older uv the duration silently drops the whole [tool.uv] table (warning only) hence losing the cooldown without failing. Dependabot's cooldown is version-independent

arnewouters (approved):

PR #405, opened 2026-09-04
fix: report non-2xx upstream responses to library mode callers

rshevchuk-git:
## Summary

In library mode (`aws_iam_streamablehttp_client`), a non-2xx HTTP response from the endpoint was never delivered to the caller. This was reported against a WAF-protected endpoint that answers oversize request bodies with `413` (empty body, no `content-type`): the tool call never returned. Subprocess mode is unaffected, because `ToolErrorMiddleware` bounds and converts tool-call failures there.

### Root cause

The mcp 1.x streamable HTTP transport calls `raise_for_status()` on any POST reply other than 202/404, from inside the task it spawned for that request. The resulting `HTTPStatusError` therefore tears down the transport task group instead of reaching the request it belongs to, and the caller stays parked in `BaseSession.send_request`, which waits without a timeout by default.

mcp 1.x does have a safety net — `_receive_loop`'s `finally` sends `CONNECTION_CLOSED` to every pending request — but it never lands: `MemoryObjectSendStream.send()` begins with an anyio checkpoint that raises `CancelledError` because the scope is already cancelled, and the surrounding guard is `except Exception`, which does not catch `BaseException`. `_response_streams` is left populated.

### Changes

* `mcp_proxy_for_aws/mcp1_compat.py` — a response hook, registered only on the library path, that rewrites a failed POST into the JSON-RPC error response the waiting request needs. Same design as mcp 2.x: the same `>= 400` boundary, the same rule of answering the pending request id. The module documents that it is temporary and lists what to delete on the mcp 2.x upgrade.
* `mcp_proxy_for_aws/client.py` — one import and one hook registration.
* `pyproject.toml` / `uv.lock` — declare `httpx`, which is used directly throughout but only arrived transitively through `fastmcp` → `mcp`.

Not included: the proxy path is untouched by design, and no dependency version was bumped.

### User experience

Against an endpoint returning `413` for an oversize body, with the session held on a background event loop:

| Caller shape | Before | After |
|---|---|---|
| Bare `ClientSession` (LangChain, LlamaIndex pattern) | Call never returns | `McpError: HTTP 413 Request Entity Too Large from <url>` immediately |
| Strands `MCPClient` | `Tool execution failed: Connection to the MCP server was closed`, and the session is dead — the next call raises `MCPClientInitializationError` | `Tool execution failed: HTTP 413 Request Entity Too Large from <url>`, and the next call succeeds |

The failure now names the HTTP status and the session survives it, so a single rejected request no longer ends the session.

## Checklist

* [ ] Changes are documented

> Docs intentionally left for a follow-up decision: a README note for the programmatic-access section is drafted but not part of this PR.

* [ ] Yes
* [x] No

* 276 unit tests pass; `mcp1_compat.py` at 100% statement and branch coverage.
* The end-to-end regression test fails with `HTTPStatusError` when the hook registration is removed, and the proxy-path guard fails when the hook is wired into `create_sigv4_client`, so both are real guards rather than vacuous assertions.
* Verified against **every mcp 1.x version the pin permits** — 1.24.0 through 1.29.1, ten releases — since the fix leans on SDK internals and a dependency bump could otherwise break it silently. All pass.
* Verified against every stable httpx in the declared range: 0.27.1, 0.27.2, 0.28.0, 0.28.1.
* Reproduced and verified through a real Strands `MCPClient`, one of the documented integrations, in addition to the bare-`ClientSession` shape.

  * `test_library_client_completes_an_exchange` passes against the repo's own `simple_mcp_server` over `REMOTE_ENDPOINT_URL` with real SigV4 signing.
  * https://github.com/aws/mcp-proxy-for-aws/actions/runs/33862820799/job/100990915760

## Acknowledgment

### DIFF
--- a/mcp_proxy_for_aws/mcp1_compat.py
+++ b/mcp_proxy_for_aws/mcp1_compat.py
@@ -1,0 +1,138 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Non-2xx handling that mcp 1.x lacks. Delete this module after the move to mcp 2.x.
+
+mcp 1.x calls ``raise_for_status()`` on a POST reply from inside the task it spawned for that
+request (``mcp/client/streamable_http.py``), so a non-2xx tears the transport down instead of
+answering the request. The caller stays parked in ``BaseSession.send_request``, which waits
+without a timeout by default, and a session held on a background event loop, the shape agent
+frameworks use, never recovers: the cancellation that should close the read stream also aborts
+the receive loop before it can hand pending requests their error.
+
+mcp 2.x answers the request id itself (``if response.status_code >= 400`` in the same file), so
+this module is only needed while the dependency chain caps mcp below 2.0: today
+``fastmcp>=3.2,<3.5`` -> ``fastmcp-slim`` -> ``mcp<2.0``. There is no fastmcp 3.5, so mcp 2.x
+arrives with fastmcp 4.0. Delete this module, ``tests/unit/test_mcp1_compat.py``, and the hook
+registration in ``client.py`` on that upgrade; none of the three survive it quietly, because
+2.x turns ``JSONRPCMessage`` into a union alias that cannot be constructed.
+
+Only the library path uses this. The proxy path needs to keep seeing ``httpx.HTTPStatusError``,
+which both ``ToolErrorMiddleware`` and ``AWSMCPProxyClient._connect`` inspect.
+"""
+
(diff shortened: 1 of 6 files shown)


mcp_proxy_for_aws/mcp1_compat.py
```diff
@@ -0,0 +1,138 @@
...
+async def _read_failure_detail(response: httpx.Response) -> str:
+    """Read what the endpoint said about the failure, capped at a length safe to pass on."""
+    try:
+        body = await response.aread()
+    except Exception as e:
+        logger.debug('Could not read body of HTTP %d response: %s', response.status_code, e)
+        return ''
+
+    detail = body.decode('utf-8', errors='replace').strip()
+    if len(detail) > _MAX_BODY_CHARS:
```
bidesh: Should we truncate the detail here? I am thinking no because as a proxy, we should just surface what is available.

rshevchuk-git: I think your callout makes sense. I can see the case for keeping the cap though because for the library-mode it becomes the JSON-RPC error `message` we hand back to the caller. Frameworks might often forward it into an LLM prompt, so an unbounded body becomes additional prompt content (of unknown length) in someone else's agent loop

bidesh (dismissed):

mcp_proxy_for_aws/mcp1_compat.py
rshevchuk-git: Removed the _MAX_BODY_CHARS

bidesh (approved):

arnewouters (approved):

PR #411, opened 2026-09-09
feat(cli): add --header for extra HTTP headers on signed requests

patrickq2:
## What

Adds `--header key=value` (repeatable) to send custom HTTP headers on every request,
exposing the `headers` parameter the library already supports.

## Why

AgentCore Gateway forwards configured request headers to its target, and since the
gateway signs with its own execution role, a header is the only channel for per-caller
context. `--metadata` can't serve this — it travels in the MCP `_meta` body field, not
as an HTTP header.

## How

- `--header` reuses `KeyValueAction`, so it matches `--metadata`'s shape
- Merged in `client_factory` **before** `create_sigv4_client`, so headers are covered
  by the SigV4 signature. Caller headers win over transport headers on collision
- Threaded through `ProfileOverrideMiddleware` so they survive a profile switch
- `RESERVED_HEADERS` lives in `sigv4_helper.py` and is enforced in both the CLI and
  `create_transport_with_sigv4`, so library callers get the same guarantee
- Supplied header names register into the existing `_sanitize_headers` path, and the
  client-creation log now sanitizes — it previously printed raw header values

## Testing

- 290 unit tests pass, 97% coverage; `ruff` and `pyright` clean; suite passes under
  `-W error::ResourceWarning`
- New tests cover parsing, repeated flags, reserved-header rejection (CLI *and*
  library), pre-signing merge, precedence, profile-switch forwarding, and log redaction
- Verified end-to-end against a live AgentCore Gateway with header forwarding enabled:
  the forwarded header reached the target and produced correct per-user authorization

## Checklist

* [x] Changes are documented

* [ ] Yes
* [x] No

## Acknowledgment


README.md
```diff
@@ -306,6 +307,41 @@ mcp_client = aws_iam_streamablehttp_client(
...
+    aws_region=region,
+    aws_service=service,
+    headers={'x-tenant-id': 'acme'},
+)
+```
+
+Headers the transport sets itself (`accept`, `content-type`, `mcp-session-id`) can be
+overridden, but doing so may break the connection — streamable HTTP requires specific
+`accept` and `content-type` values.
+
```
arnewouters, line 344: I think the addition to the table higher up is sufficient

```suggestion
```

mcp_proxy_for_aws/utils.py
```diff
@@ -87,22 +93,37 @@ def create_transport_with_sigv4(
...
         skip_auth: Whether to skip signing when credentials are unavailable
+        headers: Additional HTTP headers to send on every request
 
     Returns:
         StreamableHttpTransport instance with SigV4 authentication
+
+    Raises:
+        ValueError: If a header name would be overwritten by SigV4 signing
     """
+    caller_headers = headers
```
arnewouters, line 104: Not sure why we need this assignment

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -85,6 +85,36 @@ def _generate_assume_role_name(self):
...
+# receive rather than being logged in the clear at DEBUG.
+_EXTRA_SENSITIVE_HEADERS: set[str] = set()
+
+
+def register_sensitive_headers(names: Iterable[str]) -> None:
+    """Mark additional header names as sensitive for logging purposes.
+
+    Args:
+        names: Header names to redact in logs (matched case-insensitively)
+    """
```
arnewouters, line 115:
```suggestion
    """Mark additional header names as sensitive for logging purposes."""
```

mcp_proxy_for_aws/sigv4_helper.py
```diff
@@ -85,6 +85,36 @@ def _generate_assume_role_name(self):
...
+
+def find_reserved_headers(names: Iterable[str]) -> list[str]:
+    """Return any names that SigV4 signing would overwrite.
+
+    Args:
+        names: Header names to check (case-insensitive)
+
+    Returns:
+        The offending names, in the order given
+    """
```
arnewouters, line 99:
```suggestion
    """Return any names that SigV4 signing would overwrite."""
```

mcp_proxy_for_aws/cli.py
```diff
@@ -116,6 +118,17 @@ def parse_args():
         help='Metadata to inject into MCP requests as key=value pairs (e.g., --metadata AWS_REGION=us-west-2 KEY=VALUE)',
     )
 
+    parser.add_argument(
+        '--header',
+        nargs='*',
```
arnewouters, line 123: We have to watch out with endpoint swallowing here. Can you verify that this doesn't throw an error:

```sh
mcp-proxy-for-aws --header x-tenant=acme https://example.com/mcp
```

mcp_proxy_for_aws/cli.py
```diff
@@ -52,7 +53,8 @@ def __call__(
 
         for item in values:
             if '=' not in item:
-                parser.error(f'Metadata must be in key=value format, got: {item}')
+                flag = option_string or f'--{self.dest}'
+                parser.error(f'{flag} must be in key=value format, got: {item}')
```
arnewouters, line 57: I think this can be just
```suggestion
                parser.error(f'{option_string} must be in key=value format, got: {item}')
```

arnewouters (changes requested):

mcp_proxy_for_aws/utils.py
patrickq2: client_factory has its own headers parameter that shadows the outer one, and the inner name is fixed by the McpHttpClientFactory protocol (mcp calls it as httpx_client_factory(headers=..., ...)), so the alias is the only way to reach the caller-supplied dict inside the closure. Dropping it silently turns the merge into {**headers, **headers} and loses the custom headers entirely. I could rename the parameter to something like `extra_headers` and get rid of the assignment or leave as is, whichever you prefer.

mcp_proxy_for_aws/cli.py
patrickq2: Agreed, simplified — KeyValueAction is only attached to optionals (--metadata, --header), so argparse always passes option_string and the fallback was dead code.

mcp_proxy_for_aws/cli.py
patrickq2: Good catch, it did throw: --header must be in key=value format, got: https://example.com/mcp. nargs='*' was swallowing the endpoint positional. Dropped nargs so --header takes one key=value and is repeatable (KeyValueAction already accumulates across occurrences), which matches curl -H. The endpoint is now safe in any position; the trade-off is --header a=1 b=2 on one flag now fails loudly with unrecognized arguments: b=2 instead of silently dropping the second pair. Added tests for header-before-endpoint and the repeated form, and updated the README examples. FWIW, metadata has the same swallowing behavior today

patrickq2:
just added all changes

arnewouters (approved):

PR #412, opened 2026-09-09
docs: point readers at a real MCP server via AgentCore web search

sundargthb:
## What

Adds a "Getting an MCP Server URL" section to the root README, plus a one-line pointer to it from each of the four example READMEs.

## Why

All four examples in `examples/mcp-client/` require `MCP_SERVER_URL`, every `.env.example` uses a placeholder host, and "Example MCP Server URL Formats" gives the URL shape but never says how to get a server. A reader with no existing MCP server runs `uv run main.py` and stops there. The section gives them a working IAM-authenticated endpoint using the AgentCore CLI and the managed web search connector, which is the shortest path to a server that speaks MCP over SigV4 and has a real tool on it.

The recipe is deliberately in the root README rather than a fifth example directory. The existing four examples vary along one axis, the framework, so a use-case-named directory would have duplicated `langchain/main.py` with a different prompt string.

## Verification

Run end to end against a live gateway in `us-east-1` before opening this.

The three CLI commands run as written. `agentcore deploy` reports the endpoint as the `GatewayWsGwUrlOutput` output, and the documented `jq` expression returns the same URL from `deployed-state.json`. The gateway comes up `READY` with `authorizerType NONE`, so SigV4 is the auth and `aws_iam_streamablehttp_client` is the right client for it.

Pointing the proxy at that endpoint, `tools/list` returns `ws1___WebSearch` and `x_amz_bedrock_agentcore_search`, and `tools/call` on `ws1___WebSearch` returns results carrying `title` and `url`. Both tool names are named in the new section so nobody thinks an extra tool means a misconfigured target.

Everything created for the test was torn down.


anasstahr:
Hi, thank you for your PR! We don't think this documentation should be part of the proxy, this is more about agentcore and getting the URL itself should be intuitive

PR #414, opened 2026-09-11
feat: optionally return the backend MCP instructions

### LINKED ISSUE #413: Backend MCP instructions are replaced by the proxy's own
### What happens

The proxy returns its own `instructions` on `initialize`, discarding whatever the backend sent. A client is told about the transport instead of about the tools.

Measured against a Bedrock AgentCore Gateway on v1.6.6:

| | `serverInfo` | `instructions` |
| --- | --- | --- |
| Backend, called directly with SigV4 | `mtp-mercell-mcp` | `Tools over ... development and operational data. All tools are read-only ... Call skills_get first for guidance ...` |
| Through `mcp-proxy-for-aws` | `MCP Proxy for AWS` | `MCP Proxy for AWS provides access to SigV4 protected MCP servers through a single interface. ...` |

### Why it matters

`instructions` is the one channel a backend has to tell a client what it is for, and it is delivered only on `initialize`. Our gateway uses it to point clients at a tool that loads platform guidance. That pointer never arrives, so the guidance is never loaded — and there is no way to tell from the client side that anything was dropped.

Tool names, descriptions and schemas all come through unchanged, so those remain a usable channel. `instructions` is the only one that does not.

### Where it comes from

`InitializeMiddleware._overwrite_init_options` already reads the backend's `InitializeResult` and copies one field out of it:

```python
fastmcp_ctx._session._init_options.capabilities = init_result.capabilities
```

Its docstring says the response will contain "the backend server's info instead of the proxy's defaults", but only `capabilities` is copied. `instructions` is a field on `InitializationOptions` already, so the seam exists.

### Suggested behaviour

Opt-in, so nothing changes for clients relying on the current text:

- `--proxy-instructions`, or `AWS_MCP_PROXY_INSTRUCTIONS=1`
- applied only when the backend actually sends instructions, so a backend without any keeps the proxy description rather than an empty field
- `serverInfo` left alone — there genuinely is a proxy in the path

Happy to open a PR; I have this working and tested locally.

### Environment

`mcp-proxy-for-aws` 1.6.6, Python 3.12, backend is a Bedrock AgentCore Gateway (`--service bedrock-agentcore`).

hp197:
Closes #413.

## What

The proxy returns its own `instructions` on `initialize` and discards the backend's. A client is told about the transport instead of about the tools, and `instructions` is delivered only on `initialize`, so there is no second chance.

`InitializeMiddleware._overwrite_init_options` already reads the backend's `InitializeResult` and copies `capabilities` out of it. Its docstring says the response will contain "the backend server's info instead of the proxy's defaults", but only that one field is copied. `instructions` is already a field on `InitializationOptions`, so this copies it as well.

## Opt-in, deliberately

- `--proxy-instructions`, or `AWS_MCP_PROXY_INSTRUCTIONS=1` (matching the existing `AWS_MCP_PROXY_PROFILES` pattern)
- **Default is unchanged**, so a client relying on the current proxy text is not broken by an upgrade
- Applied only when the backend actually sends instructions — a backend without any keeps the proxy description rather than an empty field
- `serverInfo` is left alone: there genuinely is a proxy in the path, and that should stay visible. This is about what the *tools* are for, not about who answers the connection.

## Testing

Three unit tests in `tests/unit/test_initialize_middleware.py`, modelled on the existing `test_on_initialize_overwrites_init_options_with_backend_info`: default off, enabled, and enabled with the backend sending `None` or `''`.

```
280 passed, TOTAL coverage 97%
```

Verified end to end against a live Bedrock AgentCore Gateway:

| | `instructions` returned |
| --- | --- |
| no opt-in | `MCP Proxy for AWS provides access to SigV4 protected MCP servers ...` |
| `--proxy-instructions` | `Tools over ... development and operational data ... Call skills_get first ...` |
| `AWS_MCP_PROXY_INSTRUCTIONS=1` / `true` / `yes` | same as above |
| `AWS_MCP_PROXY_INSTRUCTIONS=0` | proxy text |

`ruff format` and `ruff check` clean against the repository configuration.


arnewouters:
Fixed with #416

